PR #10, opened 2022-06-30
Add functionality to data_parser.py

chelseabright96:
-Add read_gtf method to parse gtf file to dataframe (expanding attributes column into multiple columns)

The code from https://github.com/openvax/gtfparse/blob/master/gtfparse/read_gtf.py. All the same functionality has been kept. 

-Add methods to check if gtf and fasta files are of the correct format. 

-Add get_sequence_from_annotation method which is the same as get_fasta method from utils.py in main branch


IsraMekki0 (approved):
All good!

PR #11, opened 2022-07-01
Db class

francesco-campi:
I have completed the DB_clasees and tested them. As of now they can successfully download the gtf and fasta files, create the reference_DB and Oligo_DB, write them in a file and reconstrict the oligo dictionary from a tsv file. 
Probably the tests currently are not working because I am using some functionalities of the data_parser file which haven't been pushed yet, but I have saved the output I obtained in the notebook.
Moreover, the `create_DB` functions are protected in the base class, in fact within the base class they shouldn't be called since we do not have a fasta and gtf file. They can only be called from the NCBI, Ensembl, and custom classes.

The filtering functionalities are not implemented yet, but I will add them with also the first filters so that I can also test them.

Finally, I had a couple of ideas about how to make the code more efficient, let me know what you think-
- the `gene_transcript` method could take in input also a list of genes and create the transcriptome only for those genes.  So in the case of the oligos creation, we only create the transcriptome annotation we actually need.
- in the filtering of the probes, the additional features (e.g. Tm, gc cont, ...) were computed for each sequence, even if were computed also previously for a probe with the same sequence but with a different location. Since the filters use only the sequence to compute the additional features, would It make sense to compute these features only once for all the duplicates of a probe to avoid repeating the same computation?

### DIFF
--- a/oligo_designer_toolsuite/IO/_database.py
+++ b/oligo_designer_toolsuite/IO/_database.py
@@ -1,25 +1,32 @@
-import pandas as pd
 import os
-import warnings
 import random
 import shutil
+import warnings
 from pathlib import Path
 
-import pyfaidx
+import _data_parser as data_parser
+import _ftp_loader as ftp_loader
 import gtfparse
+import pandas as pd
+import pyfaidx
 from Bio import SeqIO
 
-import _ftp_loader as ftp_loader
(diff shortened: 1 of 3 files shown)


IsraMekki0 (approved):
Makes sense to me, thanks!

PR #12, opened 2022-07-11
Pre filters

francesco-campi:
I have implemented the filters we already had in `annotation_loader` and now we firstly create all the possible probes for the genes and then filter them according to the given filters. All the filter functionalities are in the `Oligos` class.


oligo_designer_toolsuite/oligo_pre_filter/_filter_base.py
```diff
@@ -1,16 +1,130 @@
+import warnings
 
+from Bio.SeqUtils import GC
+from Bio.SeqUtils import MeltingTemp as mt
 
-def _filter_by_masked_sequences():
+
+class Filter:
```
IsraMekki0, line 7: Better make it an abstract class here (inherits from ABC)

oligo_designer_toolsuite/oligo_pre_filter/_filter_base.py
```diff
@@ -1,16 +1,130 @@
...
+    """Base class that gives the structure."""
+
+    def __init__(self) -> None:
         pass
 
-def _filter_by_GC_content():
-    pass
+    def apply(self, sequence):
```
IsraMekki0, line 13: abstract method

oligo_designer_toolsuite/oligo_transcript_generation/_oligos.py
```diff
@@ -0,0 +1,285 @@
...
+                                gene_probes[gene_id][probe_sequence] = {
+                                    "transcript_id": [transcript_id],
+                                    "exon_id": [exon_id],
+                                    "chromosome": chrom,
+                                    "start": [probe_start],
+                                    "end": [probe_end],
+                                    "strand": strand,
+                                }
+
+            print(f"the total number of probes found: {loaded_probes}")
```
IsraMekki0, line 209: Not sure if we want to use print. @lisa-sousa ?

IsraMekki0 (approved):
Looks good to me, thank you

PR #13, opened 2022-07-21
Unit tests

francesco-campi:
Hi! I have implemented the tests for the DB class and the pre-filters. In particular, I checked that:

- pre-filters return the correct output
- the gene transcript created is the same as in `annotation_loader`
- under the same condition the list of sequences created is the same as in `annotation_loader`
- the oligo_DB dictionary remains unchanged after being written in a file and read

All the tests have been passed.


francesco-campi:
Hi! As we discussed I added the possibility to write the oligos DB as a gtf file and tested this functionality (added a couple of unit tests in `_test_DB.py`). However, to do this I had to create the probes ids after the filtering (before we created them after the specificity filters).

oligo_designer_toolsuite/IO/_database.py
```diff
@@ -112,104 +118,278 @@ def read_reference_DB(self, file_reference_DB):
...
+        columns_df.remove("seqname")
+        columns_df.remove("source")
+        columns_df.remove("feature")
+        columns_df.remove("score")
+        columns_df.remove("frame")
+        columns_df.remove("gene_id")
+        additional_columns = [
+            column for column in columns_df if (column not in columns_fixed)
+        ]
+        # crated the oligos db
```
lisa-sousa, line 173: @IsraMekki0 do you think it would make sense to move those parts to the _data_parser.py file?

IsraMekki0: Yes right, it would be cleaner. And in that case I think it's better to have only one function `read_oligos_DB` with a param `format` (tsv or gtf) in `_database.py` but 2 separate functions in `_data_parser.py`. Same for `write_oligos_DB`. Does that make sense to you @francesco-campi ?

IsraMekki0 (approved):
Looks good to me

PR #14, opened 2022-08-18
Joblib

chelseabright96:
-Added blast and bowtie filters that inherit from base filter class. Parallelization changed from multiprocessing to joblib package.
-All filters have apply function that takes as input a dictionary of probe information for all probes for user defined list of genes.


oligo_designer_toolsuite/IO/_database.py
```diff
@@ -112,196 +123,162 @@ def read_reference_DB(self, file_reference_DB):
...
+        self,
+        format,
+        file_oligos_DB_tsv=None,
+        file_oligos_DB_gtf=None,
+        file_oligos_DB_fasta=None,
+    ):
+        """Create the oligo db dictionary from a file. It can take both a tsv file of a gtf and fasta file and the format of file to process is defined by <format>.
+
+        :param format: format of file to process
+        :type format: str
```
lisa-sousa, line 136: Give options here in bracktes {'gtf', 'tsv'}

oligo_designer_toolsuite/IO/_database.py
```diff
@@ -112,196 +123,162 @@ def read_reference_DB(self, file_reference_DB):
...
+        else:
+            raise ValueError(f"{format} not recognized as a format!")
+
+    def write_oligos_DB(self, format, dir_oligos_DB=None):
+        """Writes the data structure self.oligos_DB in a file in the <file_oligos_DB_*> path.
+        The fromat of teh file is defined by <format>. <file_oligos_DB_tsv> is the sub-diretory of dir_output where the file will be written,
+        if None it will be set as the dir_annotation.
+
+        :param format: format of file to write
+        :type format: str
```
lisa-sousa, line 165: Give format options. {'gtf', 'tsv'}

oligo_designer_toolsuite/IO/_database.py
```diff
@@ -112,196 +123,162 @@ def read_reference_DB(self, file_reference_DB):
...
-                )  # not required anymore in this case
-                files_fasta.append(file_gene_transcript_fasta)
-            if gene_CDS:
+            elif region == "gene_CDS":
                 # generate gene cds
                 warnings.warn("Gene CDS not implemented yet")
-            return files_fasta
+            else:
+                raise ValueError(
+                    f"The given region does not exists. You selected {region}"
```
lisa-sousa, line 248: specify which regions are possible, so the user knows.

oligo_designer_toolsuite/IO/_database.py
```diff
@@ -383,9 +355,10 @@ def __init__(
 
         genome_assembly = "GRCh38"
```
lisa-sousa, line 356: genome assembly should not be harcoded, since it depends on the species and annotation release and should be returned correctly

oligo_designer_toolsuite/oligo_specificity_filter/_filter_base.py
```diff
@@ -0,0 +1,170 @@
...
+        :type dir_output: str
+        :param dir_annotations: directory storing file annotations of probes, defaults to None
+        :type dir_annotations: str, optional
+        :param number_subbatches: number of subbatches to split batched data into to control how many genes are in one batch, defaults to None
+        :type number_subbatches: int, optional
+        :param max_genes_in_batch: max number of genes allowed in a batch, defaults to 300
+        :type max_genes_in_batch: int, optional
+        """
+        self.dir_output = dir_output
+        self.max_genes_in_batch = max_genes_in_batch  # if more than 300 genes in one batch split into subbatches to reduce required memory for loading blast results
```
lisa-sousa, line 40: as discussed, it would be nicer to have one gene = one batch

oligo_designer_toolsuite/oligo_specificity_filter/_filter_base.py
```diff
@@ -0,0 +1,170 @@
...
+        self.duplicated_sequences = None
+        self.number_subbatches = number_subbatches
+
+        if n_jobs == None:
```
lisa-sousa, line 44: usually n_job=-1 means take all cores

oligo_designer_toolsuite/oligo_specificity_filter/_filter_base.py
```diff
@@ -0,0 +1,170 @@
...
+            self.n_jobs = joblib.cpu_count()
+        else:
+            self.n_jobs = n_jobs
+
+        # set logger
+        log_file = os.path.join(self.dir_output, "logs/filter_probes.log")
```
lisa-sousa, line 50: we should think about this logging situation. do we need multiple loggers? If yes we should write a seperate function for logger setup

oligo_designer_toolsuite/oligo_specificity_filter/_filter_base.py
```diff
@@ -0,0 +1,170 @@
...
+
+    @abstractmethod
+    def apply(self, probe_info):
+        """Apply filter to list of all possible probes in probe_info dictionary given user-specified genes and save output in dir_output
+
+        :param probe_info: probe info of user-specified genes
+        :type probe_info: dict
+        """
+
+    def create_batches(self, probe_info):
```
lisa-sousa, line 73: if we only take one gene per batch, this method could be simplified

oligo_designer_toolsuite/oligo_specificity_filter/_filter_base.py
```diff
@@ -0,0 +1,170 @@
...
+        """Create batches of files that subdivide the probes and probe info such that all probes of a gene are in one batch. These batches are then used as input to the filters. Number of batches is equal to n_jobs.
+
+        :param probe_info: probe info of user-specified genes
+        :type probe_info: dict
+        """
+
+        write_oligos_DB_tsv(probe_info, self.dir_annotations + "/tmp.tsv")
```
lisa-sousa, line 80: this part I don't get. Why do we write, read and then delete? Instead we should use the dictionary database, no?

oligo_designer_toolsuite/oligo_specificity_filter/_filter_blastn.py
```diff
@@ -0,0 +1,289 @@
...
+        n_jobs,
+        file_transcriptome_fasta,
+        dir_output,
+        dir_annotations,
+        word_size,
+        percent_identity,
+        coverage,
+        probe_length_min,
+        probe_length_max,
+        ligation_region=0,
```
lisa-sousa, line 27: we should not call it ligation region but maybe something like "seed_region", which is a region where no perfect matches are allowed. This should be given as tuples with start / end coordinates wrt the probe. We should also allow multiple of those regions per probe (e.g. if the user wants to make sure that no perfect matches are at beginning and end of the probe)

oligo_designer_toolsuite/oligo_specificity_filter/_filter_blastn.py
```diff
@@ -0,0 +1,289 @@
...
+        self.probe_length_max = probe_length_max
+        self.ligation_region = ligation_region
+
+        self.dir_blast = os.path.join(self.dir_output, "blast")
+        Path(self.dir_blast).mkdir(parents=True, exist_ok=True)
+
+        self.dir_probes = os.path.join(self.dir_output, "probes_blast")
+        Path(self.dir_probes).mkdir(parents=True, exist_ok=True)
+
+        self.file_removed_genes = os.path.join(
```
lisa-sousa, line 62: I think it would be more convenient if we don't write this into a file but store it in the database. We then can have a function that writes all genes without any probes from the database.

oligo_designer_toolsuite/oligo_specificity_filter/_filter_blastn.py
```diff
@@ -0,0 +1,289 @@
...
+
+                cmd = NcbiblastnCommandline(
+                    query=file_probe_fasta_batch,
+                    db=self.file_transcriptome_fasta,
+                    outfmt="10 qseqid sseqid length qstart qend qlen",
+                    out=file_blast_batch,
+                    strand="plus",
+                    word_size=self.word_size,
+                    perc_identity=self.percent_identity,
+                    num_threads=self.n_jobs,
```
lisa-sousa, line 93: I'm not sure if we can have n_jobs in the blast command line call and also when calling this function... this would mean we will start n_job x n_jobs processes, no?

oligo_designer_toolsuite/oligo_specificity_filter/_filter_blastn.py
```diff
@@ -0,0 +1,289 @@
...
+                )
+                out, err = cmd()
+
+        # create blast database
+        self.logging.info("Creating blast database")
+        start_time = time.perf_counter()
```
lisa-sousa, line 99: remove timing. this should be seperate when we benchmark but not in the main code

oligo_designer_toolsuite/oligo_specificity_filter/_filter_blastn.py
```diff
@@ -0,0 +1,289 @@
...
+        :type probe_length_max: int
+        :param ligation_region: coverage between probes and target sequence should not span region around ligation site (e.g. ligation_region = 5 would correspond to -4 to +5 nt around ligation site), if ligation_region = 0, omit this requirement
+        :type ligation_region: int
+        """
+        super().__init__(n_jobs, dir_output, dir_annotations)
+
+        self.word_size = word_size
+        self.percent_identity = percent_identity
+        self.coverage = coverage
+        self.file_transcriptome_fasta = file_transcriptome_fasta
```
lisa-sousa, line 51: this is not necessarily the transcriptome, but background database file

oligo_designer_toolsuite/oligo_specificity_filter/_filter_blastn.py
```diff
@@ -0,0 +1,289 @@
...
+                for batch_id in range(self.n_jobs)
+            )
+
+        finish_time = time.perf_counter()
+
+        self.logging.info(
+            f"Blast results processed in {finish_time-start_time} seconds"
+        )
+
+        _write_removed_genes(self)
```
lisa-sousa, line 266: just remove all filtered probes from the database and keep genes with 0 probes, which can be written later from the database to a file

oligo_designer_toolsuite/oligo_specificity_filter/_filter_blastn.py
```diff
@@ -0,0 +1,289 @@
...
+
+        # remove intermediate files
+        shutil.rmtree(self.dir_blast)
+        for file in os.listdir(self.dir_annotations):
+            if re.search("probes_*", file):
+                os.remove(os.path.join(self.dir_annotations, file))
+
+        for file in os.listdir(self.dir_annotations):
+            if re.search(
+                "reference_DB_unknown_unknown_Custom_release_unknown_genome_False_gene_transcript_True.n*",
```
lisa-sousa, line 276: why is that hardcoded here?

oligo_designer_toolsuite/oligo_specificity_filter/_filter_blastn.py
```diff
@@ -0,0 +1,289 @@
...
+        """Process the output of the BlastN alignment search"""
+
+        def _process_blast_results(batch_id):
+
+            probes_info = _load_probes_info(self, batch_id)
+
+            num_probes_wo_match = 0
+            for subbatch_id in range(self.number_subbatches):
+                blast_results = _read_blast_output(batch_id, subbatch_id)
+                num_probes_wo_match += _filter_probes_blast(probes_info, blast_results)
```
lisa-sousa, line 129: we should not need that anymore when we simply delete the probes from the probe database, like in the other filters

oligo_designer_toolsuite/oligo_specificity_filter/_filter_blastn.py
```diff
@@ -0,0 +1,289 @@
...
+        self.logging.info(
+            f"Blast alignmnet search finished in {finish_time-start_time} seconds"
+        )
+
+    def filter_probes_by_blast_results(self):
+        """Process the output of the BlastN alignment search"""
+
+        def _process_blast_results(batch_id):
+
+            probes_info = _load_probes_info(self, batch_id)
```
lisa-sousa, line 124: we should access this information directly from the probes database

oligo_designer_toolsuite/oligo_specificity_filter/_filter_blastn.py
```diff
@@ -0,0 +1,289 @@
...
+            blast_results_matches = blast_results_matches.merge(
+                probes_info[["probe_id", "ligation_site"]],
+                left_on="query",
+                right_on="probe_id",
+                how="inner",
+            )
+            blast_results_matches.drop(columns=["probe_id"], inplace=True)
+            blast_results_matches_filtered = []
+
+            for probe_length in range(self.probe_length_min, self.probe_length_max + 1):
```
lisa-sousa, line 192: instead of looping over the different probe length, we could add another column where we store the min_alignment_length

oligo_designer_toolsuite/oligo_specificity_filter/_filter_blastn.py
```diff
@@ -0,0 +1,289 @@
...
+                else:
+                    blast_results_matches_probe_length = (
+                        blast_results_matches_probe_length[
+                            (
+                                blast_results_matches_probe_length.alignment_length
+                                > min_alignment_length
+                            )
+                            | (
+                                (
+                                    blast_results_matches_probe_length.query_start
```
lisa-sousa, line 215: this part should be replaced with the see_regions tuple format. Here we could use the an interval intersection, i.e. if (start,end) of seed_region are within (start,end) of query, then we can discard this probe, because we found a perfect match covering the seed_region

oligo_designer_toolsuite/oligo_specificity_filter/_filter_blastn.py
```diff
@@ -0,0 +1,289 @@
...
+                                )
+                            )
+                        ]
+                    )
+
+                blast_results_matches_filtered.append(
+                    blast_results_matches_probe_length
+                )
+
+            blast_results_matches_filtered = pd.concat(blast_results_matches_filtered)
```
lisa-sousa, line 234: this part should all be integrated into the probe database, i.e. every probe that is filtered out get immediately removed from the probe database

oligo_designer_toolsuite/oligo_specificity_filter/_filter_blastn.py
```diff
@@ -0,0 +1,289 @@
...
+                ~blast_results["query"].isin(probes_with_match)
+            ]
+
+            for gene_id in blast_results["query_gene_id"].unique():
+                probes_wo_match_gene = probes_wo_match[
+                    probes_wo_match.query_gene_id == gene_id
+                ]
+                probes_wo_match_gene = probes_wo_match_gene["query"].unique()
+
+                if len(probes_wo_match_gene) > 0:  # gene has to have at least one probe
```
lisa-sousa, line 247: we don't need this filter here. Just keep genes without probes in the database

oligo_designer_toolsuite/oligo_specificity_filter/_filter_bowtie.py
```diff
@@ -0,0 +1,337 @@
...
+)
+
+
+class ProbeFilterBowtie(ProbeFilterBase):
+    def __init__(
+        self,
+        n_jobs,
+        dir_output,
+        dir_annotations,
+        file_transcriptome_fasta,
```
lisa-sousa, line 25: again, this should be background database

oligo_designer_toolsuite/oligo_specificity_filter/_filter_bowtie.py
```diff
@@ -0,0 +1,337 @@
...
+        min_mismatches=4,
+        mismatch_region=None,
+        ligation_region=0,
```
lisa-sousa, line 28: Here, again we should merge ligation_site and ligation_region variables into seed_Region, which is defined as tuple (see blast filter)

oligo_designer_toolsuite/oligo_specificity_filter/_filter_bowtie.py
```diff
@@ -0,0 +1,337 @@
...
+            raise ValueError(
+                "Choice of mismatch_region out of range for bowtie allignment tool. Please choose a value no less than 5"
+            )
+        else:
+            self.mismatch_region = mismatch_region
+
+        self.dir_bowtie = os.path.join(self.dir_output, "bowtie")
+        Path(self.dir_bowtie).mkdir(parents=True, exist_ok=True)
+
+        self.dir_probes = os.path.join(self.dir_output, "probes_bowtie")
```
lisa-sousa, line 69: we should not write anything else than intermediate files, which are deleted afterwards, and store all other information in the probes DB

oligo_designer_toolsuite/oligo_specificity_filter/_filter_bowtie.py
```diff
@@ -0,0 +1,337 @@
...
+        Path(self.dir_probes).mkdir(parents=True, exist_ok=True)
+
+        self.file_removed_genes = os.path.join(
```
lisa-sousa, line 72: again, store this info in probes DB

oligo_designer_toolsuite/oligo_specificity_filter/_filter_bowtie.py
```diff
@@ -0,0 +1,337 @@
...
+                        + " "
+                        + file_bowtie_batch
+                    )
+
+                process = Popen(command, shell=True, cwd=self.dir_bowtie).wait()
+
+        index_exists = False
+        # Check if bowtie index exists
+        for file in os.listdir(self.dir_bowtie):
+            if re.search("^transcriptome*", file):
```
lisa-sousa, line 126: should not be called transcriptome since it can be different regions -> it's the background DB instead

oligo_designer_toolsuite/oligo_specificity_filter/_filter_bowtie.py
```diff
@@ -0,0 +1,337 @@
...
+            process = Popen(command1, shell=True, cwd=self.dir_bowtie).wait()
+
+        start_time = time.perf_counter()
+
+        with parallel_backend("loky"):
+            Parallel()(
+                delayed(_run_bowtie_batch)(batch_id) for batch_id in range(self.n_jobs)
+            )
+
+        finish_time = time.perf_counter()
```
lisa-sousa, line 144: No time logging in main code

oligo_designer_toolsuite/oligo_specificity_filter/_filter_bowtie.py
```diff
@@ -0,0 +1,337 @@
...
+                    "probes_sequence_batch{}_{}.fna".format(batch_id, subbatch_id),
+                )
+                file_bowtie_batch = os.path.join(
+                    self.dir_bowtie,
+                    "bowtie_batch{}_{}.txt".format(batch_id, subbatch_id),
+                )
+
+                if self.mismatch_region is not None:
+                    command = (
+                        "bowtie transcriptome -f -a -n "
```
lisa-sousa, line 102: don't call it transcriptome

oligo_designer_toolsuite/oligo_specificity_filter/_filter_bowtie.py
```diff
@@ -0,0 +1,337 @@
...
+            if re.search("bowtie_*", file):
+                os.remove(os.path.join(self.dir_bowtie, file))
+
+        for file in os.listdir(self.dir_annotations):
+            if re.search("probes_*", file):
+                os.remove(os.path.join(self.dir_annotations, file))
+
+        for file in os.listdir(self.dir_annotations):
+            if re.search(
+                "reference_DB_unknown_unknown_Custom_release_unknown_genome_False_gene_transcript_True.n*",
```
lisa-sousa, line 325: why is that hardcoded? this information should be saved in a variable

oligo_designer_toolsuite/oligo_specificity_filter/_filter_bowtie.py
```diff
@@ -0,0 +1,337 @@
...
+            f"Bowtie alignmnet search finished in {finish_time-start_time} seconds"
+        )
+
+    def filter_probes_by_bowtie_results(self):
+        """Use bowtie results to filter probes based on number of mismatches and mismatch position
+
+        :return: Text file for each gene containing the filtered probes for that gene. Additionally, a text file containing a list of genes with insufficient number of probes is returned.
+        :rtype: Txt files"""
+
+        def _process_bowtie_results(batch_id):
```
lisa-sousa, line 155: this function looks quite familiar to the process_blast_results function. Can we merge this?

oligo_designer_toolsuite/oligo_specificity_filter/_filter_bowtie.py
```diff
@@ -0,0 +1,337 @@
...
+
+            probes_info = _load_probes_info(self, batch_id)
+
+            num_probes_wo_match = 0
```
lisa-sousa, line 159: why do we store this information? It is not used anywhere, no?

oligo_designer_toolsuite/oligo_specificity_filter/_filter_bowtie.py
```diff
@@ -0,0 +1,337 @@
...
+                bowtie_results["query_gene_id"] != bowtie_results["reference_gene_id"]
+            ]
+
+            if self.ligation_region > 0:
+
+                # Get probes where mismatch positions are given
+                bowtie_results_matches_mismatch_positions = bowtie_results_matches[
+                    bowtie_results_matches["mismatch_positions"].notna()
+                ]
+                # Extract positions (Positions of mismatches given in bowtie output)
```
lisa-sousa, line 230: I'm not sure if this is correct because we don't consider if the match is on the + or - strand. If it's on the minus strand the missmatches should be counted on the reverse complement

oligo_designer_toolsuite/oligo_specificity_filter/_filter_bowtie.py
```diff
@@ -0,0 +1,337 @@
...
+                )
+
+                bowtie_results_add_ligation = bowtie_results_positions.merge(
+                    probes_info[["probe_id", "ligation_site"]],
+                    left_on="query",
+                    right_on="probe_id",
+                    how="inner",
+                )
+
+                # Calculate ligation region
```
lisa-sousa, line 248: this should be done with (start,end) intervals again

oligo_designer_toolsuite/oligo_specificity_filter/_filter_exact_matches.py
```diff
@@ -0,0 +1,203 @@
...
+from Bio.SeqRecord import SeqRecord
+from joblib import Parallel, delayed, parallel_backend
+
+from oligo_designer_toolsuite.IO._data_parser import read_oligos_DB_tsv
+
+from ._filter_base import ProbeFilterBase
+
+
+class ProbeFilterExact(ProbeFilterBase):
+    def __init__(self, n_jobs, dir_output, dir_annotations):
```
lisa-sousa, line 17: this whole class should be rewritten such that we do not operate with reading / writing files but instead directly on the probe DB -> we can call this line ` duplicated_sequences = list(iteration_utilities.unique_everseen(iteration_utilities.duplicates(sequences)))` directly on all sequences of the probe DB and then filter out the returned sequences.

oligo_designer_toolsuite/oligo_specificity_filter/_filter_bowtie2.py
```diff
@@ -0,0 +1,280 @@
...
+from ._utils import _load_probes_info, _write_output, _write_removed_genes
+
+
+class ProbeFilterBowtie2(ProbeFilterBase):
+    def __init__(
+        self,
+        n_jobs,
+        dir_output,
+        dir_annotations,
+        file_transcriptome_fasta,
```
lisa-sousa, line 20: should be background DB not transcriptome

oligo_designer_toolsuite/oligo_specificity_filter/_filter_bowtie2.py
```diff
@@ -0,0 +1,280 @@
...
+        :param min_score: User defined threshhold for alignmnent score. If specified, the Bowtie 2 filter filters out probes with an aligment score greater than min_score. If None, min_score defaults to -0.6 + -0.6 * L, where L is the read length
+        """
+
+        super().__init__(n_jobs, dir_output, dir_annotations)
+
+        self.n_jobs = n_jobs
+        self.file_transcriptome_fasta = file_transcriptome_fasta
+        self.min_score = min_score
+
+        self.dir_bowtie2 = os.path.join(self.dir_output, "bowtie2")
```
lisa-sousa, line 39: we should only write files if necessary (e.g. temp files) and then remove them. Everything else should be stored in probes DB

oligo_designer_toolsuite/oligo_specificity_filter/_filter_bowtie2.py
```diff
@@ -0,0 +1,280 @@
...
+                        + " -S "
+                        + file_bowtie2_batch
+                    )
+
+                process = Popen(command, shell=True, cwd=self.dir_bowtie2).wait()
+
+        # Check if bowtie2 index exists
+        index_exists = False
+        for file in os.listdir(self.dir_bowtie2):
+            if re.search("^transcriptome_bowtie2*", file):
```
lisa-sousa, line 97: should not be called transcriptome but background DB

oligo_designer_toolsuite/oligo_specificity_filter/_filter_bowtie2.py
```diff
@@ -0,0 +1,280 @@
...
+        # Create bowtie2 index if none exists
+        if not index_exists:
+            command1 = (
+                "bowtie2-build -f "
+                + self.file_transcriptome_fasta
+                + " transcriptome_bowtie2"
+            )
+            process = Popen(command1, shell=True, cwd=self.dir_bowtie2).wait()
+
+        start_time = time.perf_counter()
```
lisa-sousa, line 110: no timing in main code

oligo_designer_toolsuite/oligo_specificity_filter/_filter_bowtie2.py
```diff
@@ -0,0 +1,280 @@
...
+        """Use bowtie 2 results to filter probes based on min_score parameter
+
+        :return: Text file for each gene containing the filtered probes for that gene. Additionally, a text file containing a list of genes with insufficient number of probes is returned.
+        :rtype: Txt files"""
+
+        def _process_bowtie2_results(batch_id):
+
+            probes_info = _load_probes_info(self, batch_id)
+
+            num_probes_wo_match = 0
```
lisa-sousa, line 132: I don't see that we use num_probe_wo_match anywhere, please remove

oligo_designer_toolsuite/oligo_specificity_filter/_filter_bowtie2.py
```diff
@@ -0,0 +1,280 @@
...
+
+    def filter_probes_by_bowtie2_results(self):
+        """Use bowtie 2 results to filter probes based on min_score parameter
+
+        :return: Text file for each gene containing the filtered probes for that gene. Additionally, a text file containing a list of genes with insufficient number of probes is returned.
+        :rtype: Txt files"""
+
+        def _process_bowtie2_results(batch_id):
+
+            probes_info = _load_probes_info(self, batch_id)
```
lisa-sousa, line 130: this info should be taken from probes DB

oligo_designer_toolsuite/oligo_specificity_filter/_filter_blastn.py
```diff
@@ -0,0 +1,289 @@
...
+        )
+
+    def filter_probes_by_blast_results(self):
+        """Process the output of the BlastN alignment search"""
+
+        def _process_blast_results(batch_id):
+
+            probes_info = _load_probes_info(self, batch_id)
+
+            num_probes_wo_match = 0
```
lisa-sousa, line 126: don't see that we use num_probes_wo_match anywhere -> remove

oligo_designer_toolsuite/oligo_specificity_filter/_filter_bowtie.py
```diff
@@ -0,0 +1,337 @@
...
+
+    def filter_probes_by_bowtie_results(self):
+        """Use bowtie results to filter probes based on number of mismatches and mismatch position
+
+        :return: Text file for each gene containing the filtered probes for that gene. Additionally, a text file containing a list of genes with insufficient number of probes is returned.
+        :rtype: Txt files"""
+
+        def _process_bowtie_results(batch_id):
+
+            probes_info = _load_probes_info(self, batch_id)
```
lisa-sousa, line 157: this info should be taken from probes DB

oligo_designer_toolsuite/oligo_specificity_filter/_filter_bowtie2.py
```diff
@@ -0,0 +1,280 @@
...
+                ]
+                probes_wo_match_gene = probes_wo_match_gene["probe_id"].unique()
+
+                if len(probes_wo_match_gene) > 0:  # gene has to have at least one probe
+                    _write_output(self, probes_info, gene_id, probes_wo_match_gene)
+
+            return len(probes_wo_match["probe_id"].unique())
+
+        # Parallelize filtering of probes
+        start_time = time.perf_counter()
```
lisa-sousa, line 242: no timing in main code

oligo_designer_toolsuite/oligo_specificity_filter/_filter_bowtie2.py
```diff
@@ -0,0 +1,280 @@
...
+            :type blast_results: pandas.DataFrame
+            """
+
+            bowtie2_results_matches = bowtie2_results[
+                bowtie2_results["query_gene_id"] != bowtie2_results["reference_gene_id"]
+            ]
+
+            probes_with_match = bowtie2_results_matches["query"].unique()
+            probes_wo_match = probes_info[
+                ~probes_info["probe_id"].isin(probes_with_match)
```
lisa-sousa, line 227: the results should be directly applied to probes DB

oligo_designer_toolsuite/oligo_specificity_filter/_utils.py
```diff
@@ -0,0 +1,81 @@
+import os
+
+import pandas as pd
+
+
+def _write_output(self, probes_info, gene_id, probes_wo_match):
```
lisa-sousa, line 6: we dn't need that function if we plly the filtering directly on probes DB. But we should add an additional function in I/O to write genes without probes into a file (or genes with probes < x)

oligo_designer_toolsuite/oligo_specificity_filter/_utils.py
```diff
@@ -0,0 +1,81 @@
...
+    :type gene_id: string
+    :param probes_wo_match: List of suitable probes that don't have matches in the transcriptome.
+    :type probes_wo_match: list
+    """
+    file_output = os.path.join(self.dir_probes, "probes_{}.txt".format(gene_id))
+    valid_probes = probes_info[probes_info["probe_id"].isin(probes_wo_match)]
+    valid_probes.to_csv(file_output, sep="\t", index=False)
+
+
+def _load_probes_info(self, batch_id):
```
lisa-sousa, line 20: we should retrieve this information directly from probes DB

oligo_designer_toolsuite/oligo_specificity_filter/_utils.py
```diff
@@ -0,0 +1,81 @@
...
+            "melt_temp_arm2": float,
+            "dif_melt_temp_arms": float,
+            "ligation_site": int,
+        },
+    )
+
+    return probes_info
+
+
+def _write_removed_genes(self):
```
lisa-sousa, line 55: this should become an i/O function applied to the probes DB

oligo_designer_toolsuite/oligo_specificity_filter/_utils.py
```diff
@@ -0,0 +1,81 @@
...
+        gene_id = probe_file[len("probes_") : -len(".txt")]
+        if gene_id in self.removed_genes:
+            self.removed_genes.remove(gene_id)
+
+    with open(self.file_removed_genes, "w") as output:
+        for gene_id in self.removed_genes:
+            output.write(f"{gene_id}\t0\n")
+
+
+def mismatch_in_ligation(row):
```
lisa-sousa, line 70: we don't need this function -> see comments in bowtie filter

oligo_designer_toolsuite/oligo_transcript_generation/_gene_transcript.py
```diff
@@ -23,17 +24,19 @@ def __init__(self, file_sequence, file_annotation):
             self.file_annotation
         )  # dataframe with annotation file
 
-    def generate_for_reference(self, block_size, dir_output="./output/annotation"):
+    def generate_for_reference(
+        self, block_size, file_gene_transcript_fasta, dir_out="output/annotation"
```
lisa-sousa, line 28: we should not call it `file_gene_transcript_fasta` because the background dB can be something else than transcripts

lisa-sousa (commented):
Some major comments:
1. we should not read/write files but use our database structure that we setup. Exceptions are temp files which are output from other tools, e.g. blast
2. we should parallelize over each gene otherwise it becomes too confusing
3. we should generalize the ligation site filter to a seed region filter which takes intervals into account that should not be covered by perfect matches.
4. the seed region filter is missing in bowtie2 and is not correct in bowtie filter
5. the exact match filter should be largely rewritten and adapted to our database structure

oligo_designer_toolsuite/IO/_database.py
```diff
@@ -383,9 +379,10 @@ def __init__(
...
```
lisa-sousa, line 380: genome assembly should not be harcoded, since it depends on the species and annotation release and should be returned correctly

oligo_designer_toolsuite/oligo_specificity_filter/_filter_blastn.py
```diff
@@ -0,0 +1,193 @@
...
+        filtered_oligo_DBs = Parallel(n_jobs=n_jobs)(
+            delayed(self._run_blast)(oligo_DB[gene], gene, database_name)
+            for gene in genes
+        )
+
+        # reconstruct the oligos db and return it
+        for gene, filtered_oligo_DB in zip(genes, filtered_oligo_DBs):
+            oligo_DB[gene] = filtered_oligo_DB
+        # remove the files
+        """for file in os.listdir(self.dir_blast):
```
lisa-sousa, line 86: are we not removing the blast files anymore, when the blast search is done?

oligo_designer_toolsuite/oligo_specificity_filter/_filter_bowtie2.py
```diff
@@ -0,0 +1,206 @@
...
+        filtered_oligo_DBs = Parallel(n_jobs=n_jobs)(
+            delayed(self._run_bowtie2)(oligo_DB[gene], gene, index_name)
+            for gene in genes
+        )
+
+        # reconstruct the oligos_DB
+        for gene, filtered_oligo_DB in zip(genes, filtered_oligo_DBs):
+            oligo_DB[gene] = filtered_oligo_DB
+        # remove the files
+        """for file in os.listdir(self.dir_bowtie2):
```
lisa-sousa, line 78: are we still removing the tmeporary files somewhere?

oligo_designer_toolsuite/oligo_specificity_filter/_filter_seed_region.py
```diff
@@ -0,0 +1,196 @@
...
+
+class BowtieSeedRegion(SpecificityFilterBase):
+    """This class filters probes based on the Bowtie short read alignment tool on a specific region of the probe. The region take in consideration is created according to the ``seed_region_creation`` class.
+    The user can customize the filtering by specifying the num_mismatches, and all probes with number mismatches lower or equal to min_mismatches inside the mismatch_region are filtered out.
+
+    Use conda install -c bioconda bowtie to install Bowtie package
+
+    :param file_transcriptome_fasta: path to fasta file containing all probes
+    :type file_transcriptome_fasta: str
+    :param seed_region_creation: The region of the probe where the mismatches are considered. Probes that have less than min_mismatches in the first L bases (where L is 5 or greater) are filtered out
```
lisa-sousa, line 20: "In the first L" bases might be confusing since it can also be in the middle of the probe. Maybe rephrase

lisa-sousa (changes requested):
Looks really good now. I just left a few minor comments. When they are addressed, you can merge from my side if @IsraMekki0 thinks it's fine.

francesco-campi:
I delete the fasta files and the results of the search right after having used them, so in the _run_blastn (or equivalent) function. So the database or index files are kept in case they will be needed once more.

oligo_designer_toolsuite/IO/_database.py
francesco-campi: I would modify this in the branch where we split the reference DB from the oligos DB. So w can also minimize the merge conflicts with the dev branch.

lisa-sousa: ok, sounds good

oligo_designer_toolsuite/oligo_specificity_filter/_filter_seed_region.py
lisa-sousa: Did you address this comment?

lisa-sousa (approved):
Looks all good to me now. If @IsraMekki0 thinks it's fine we can merge to dev

oligo_designer_toolsuite/IO/_data_parser.py
```diff
@@ -234,16 +237,12 @@ def parse_frame(s):
                     for s in df[fix_quotes_column]
                 ]
             dataframes.append(df)
-    except Exception as e:
```
IsraMekki0, line 237: Keep raising the exception and print the message inside `ParsingError`. Otherwise, the program would go on

oligo_designer_toolsuite/IO/_data_parser.py
```diff
@@ -234,16 +237,12 @@ def parse_frame(s):
...
-        raise ParsingError(str(e))
+    except:
+        print("An error occured when parsing the gtf file")
     df = pd.concat(dataframes)
     return df
 
 
-class ParsingError(Exception):
```
IsraMekki0, line 243: print("An error occured when parsing the gtf file")

oligo_designer_toolsuite/IO/_data_parser.py
```diff
@@ -380,11 +379,281 @@ def merge_fasta(files_fasta, file_merged_fasta):
...
+        "end",
+        "strand",
+        "length",
+    ]
+    columns_df = list(oligos_df.columns)
+    # remove the columns we don't need
+    columns_df.remove("seqname")
+    columns_df.remove("source")
+    columns_df.remove("feature")
+    columns_df.remove("score")
```
IsraMekki0, line 434: Are these columns always there? If not, this would raise an error

oligo_designer_toolsuite/IO/_data_parser.py
```diff
@@ -380,11 +379,281 @@ def merge_fasta(files_fasta, file_merged_fasta):
...
+    columns_df.remove("frame")
+    columns_df.remove("gene_id")
+    additional_columns = [
+        column for column in columns_df if (column not in columns_fixed)
+    ]
+    # crated the oligos db
+    current_gene = ""
+    current_probe = ""
+    for index in range(len(oligos_df)):
```
IsraMekki0, line 443: `for row in oligos_df` would probably be faster

oligo_designer_toolsuite/IO/_data_parser.py
```diff
@@ -380,11 +379,281 @@ def merge_fasta(files_fasta, file_merged_fasta):
...
+            oligos_DB[current_gene] = {}
+        if row["seqname"] != current_probe:
+            current_probe = row["seqname"]
+            oligos_DB[current_gene][current_probe] = {}
+            oligo_fasta = next(oligos_fasta)
+            probe_sequence = oligo_fasta.seq
+            assert (
+                oligo_fasta.id == current_probe
+            )  # check that the fasta and gtf file are in sync
+            oligos_DB[current_gene][current_probe] = {}
```
IsraMekki0, line 456: Duplication here

oligo_designer_toolsuite/IO/_data_parser.py
```diff
@@ -380,11 +379,281 @@ def merge_fasta(files_fasta, file_merged_fasta):
...
+        row = oligos_df.iloc[index]
+        if row["gene_id"] != current_gene:
+            current_gene = row["gene_id"]
+            oligos_DB[current_gene] = {}
+        if row["seqname"] != current_probe:
+            current_probe = row["seqname"]
+            oligos_DB[current_gene][current_probe] = {}
+            oligo_fasta = next(oligos_fasta)
+            probe_sequence = oligo_fasta.seq
+            assert (
```
IsraMekki0, line 453: This is not handled, does this happen?

oligo_designer_toolsuite/IO/_data_parser.py
```diff
@@ -380,11 +379,281 @@ def merge_fasta(files_fasta, file_merged_fasta):
...
+    The order of columns is : probe_id, probe_sequence, gene_id,  'transcript_id', 'exon_id', 'chromosome', 'start', 'end', 'strand', all the additional info computed by the filtering class.
+
+    :param file_oligos_DB_tsv: path of the oligos_DB file
+    :type file_oligos_DB_tsv: str
+    :return: oligos_DB
+    :rtype: dict
+    """
+
+    def parse_line_tsv(line, oligos_DB, current_gene, add_features):
+        """Parses the lines of the tsv file of the oligos db and puts teh data in the dictionary.
```
IsraMekki0, line 492: ..puts *the*...

oligo_designer_toolsuite/oligo_pre_filter/_filter_base.py
```diff
@@ -14,6 +14,7 @@ def __init__(self) -> None:
     def apply(self, sequence):
         """Applies teh filters and returns if the sequence fulfillts the constraints and the additional features computed in a dictionary.
```
IsraMekki0, line 15: Typos:
... reimplemented in **the** ...
The **additional** computed features...

oligo_designer_toolsuite/IO/_database.py
```diff
@@ -112,196 +132,162 @@ def read_reference_DB(self, file_reference_DB):
...
+                file_oligos_DB_gtf, file_oligos_DB_fasta
+            )
+            self.file_oligos_DB_gtf = file_oligos_DB_gtf
+            self.file_oligos_DB_fasta = file_oligos_DB_fasta
+        else:
+            raise ValueError(f"{format} not recognized as a format!")
+
+    def write_oligos_DB(self, format, dir_oligos_DB=None):
+        """Writes the data structure self.oligos_DB in a file in the ``file_oligos_DB_*`` path.
+        The fromat of teh file is defined by ``format``. ``file_oligos_DB_tsv`` is the sub-diretory of dir_output where the file will be written,
```
IsraMekki0, line 170: The **format** of **the** file is defined..

oligo_designer_toolsuite/IO/_database.py
```diff
@@ -112,196 +132,162 @@ def read_reference_DB(self, file_reference_DB):
...
-        :param genome: create the reference file for the whole genome, defaults to False
-        :type genome: bool, optional
-        :param gene_transcript: create the reference file for the gene transcript, defaults to True
-        :type gene_transcript: bool, optional
-        :param gene_CDS: create the reference file for the coding region, defaults to False
-        :type gene_CDS: bool, optional
-        :param dir_output: folder where the fasta file will be saved, defaults to './output/annotation'
-        :type dir_output: str, optional
-        :param block_size: size of the exon junctions, defaults to None
+        """Creates a fasta file for each of the region selected (genome, gene_transcript, gene_CDS) which will be used for alignements, default is "gene_transcript".
```
IsraMekki0, line 216: ...specified the exon **junctions** size is set to...
...**subdirectory** of dir_out where...

oligo_designer_toolsuite/IO/_database.py
```diff
@@ -112,196 +132,162 @@ def read_reference_DB(self, file_reference_DB):
...
         genes=None,
         region="gene_transcript",
-        number_batchs=1,
-        dir_output="./output/annotation",
-        write=True,
+        n_jobs=2,
     ):
-        """creates the DB containing all the oligo sequence extracted form the given <region> and belonging the the specified genes. If no genes are specified then
-        will be used all the genes. The DB is a dictionary data structure and can be written in a tsv format by setting <write> = True.
+        """creates the DB containing all the oligo sequence extracted form the given ``region`` and belonging the the specified genes. If no genes are specified then
```
IsraMekki0, line 282: Should state here that is does not automatically write

oligo_designer_toolsuite/IO/_database.py
```diff
@@ -325,23 +311,33 @@ def create_target_region(region, genes):
...
+            file_region_annotation, genes, n_jobs, self.dir_annotation
         )
-        if write:
-            self.write_oligos_DB()
-
         # clean folder
         os.remove(file_region_annotation)
 
+    def remove_genes_with_insufficient_probes(self, pipeline_step, write=True):
+        """Deletes from the `oligo_DB` the genes which do not have any more probes, and writes them in a file with the name of the step of teh pipeline at which they have been deleted.
```
IsraMekki0, line 327: ...name of the step of **the** pipeline..

IsraMekki0, line 327: ...and **optionally** writes them in a file with the name of the step of **the** pipeline at...

oligo_designer_toolsuite/oligo_specificity_filter/_filter_base.py
```diff
@@ -0,0 +1,58 @@
...
+        """Construnctor"""
+
+        self.dir_specificity = (
+            dir_specificity  # folder where we write the intermediate files
+        )
+        os.makedirs(self.dir_specificity, exist_ok=True)
+
+    @abstractmethod
+    def apply(self, oligo_DB, file_reference_DB, n_jobs):
+        """Apply filter to list of all possible probes in probe_info dictionary and filter out the probes which don't fulfill the requiremetns.
```
IsraMekki0, line 25: ...filter out the probes which don't **fulfil** the **requirements**.
        :param oligo_DB: database containing the probes **and** their features

oligo_designer_toolsuite/oligo_specificity_filter/_filter_blastn.py
```diff
@@ -0,0 +1,189 @@
...
+        :return: probe info of user-specified genes
+        :rtype : dict
+        """
+
+        # create blast database
+        database_exists = False
+        database_name = os.path.basename(file_reference_DB)
+        # Check if blast database exists
+        for file in os.listdir(self.dir_blast):
+            if re.search(f"^{database_name}.*", file):
```
IsraMekki0, line 64: Are you missing a `database_exists = True` here?

oligo_designer_toolsuite/oligo_specificity_filter/_filter_blastn.py
```diff
@@ -0,0 +1,189 @@
...
+
+        return oligo_DB
+
+    def _run_blast(self, gene_DB, gene, database_name):
+        """Run BlastN alignment tool to find regions of local similarity between sequences, where sequences are probes and background sequences (e.g. transcript, genome, etc.).
+        BlastN identifies the transcript regions where probes match with a certain coverage and similarity.
+
+        :param gene_DB: database containing the probes form one gene
+        :type gene_DB: dict
+        :param gene: id of thet gene processed
```
IsraMekki0, line 94: ...**the** gene processed

oligo_designer_toolsuite/oligo_specificity_filter/_filter_blastn.py
```diff
@@ -0,0 +1,189 @@
...
+        file_blast_gene = os.path.join(self.dir_blast, f"blast_{gene}.txt")
+        cmd = NcbiblastnCommandline(
+            query=file_probe_fasta_gene,
+            db=os.path.join(self.dir_blast, database_name),
+            outfmt="10 qseqid sseqid length qstart qend qlen",
+            out=file_blast_gene,
+            strand="plus",
+            word_size=self.word_size,
+            perc_identity=self.percent_identity,
+            num_threads=1,  # ????
```
IsraMekki0, line 108: Good point! Need to do some benchmarking here. Last time, I tested with multiprocessing (tried several combination of `n_jobs`/ `n_threads` for the same total number of jobs). My conclusion was (oppositely to what I read everywhere) that it was faster to increase the number of jobs with `num_threads = 1`. To be confirmed with joblib

oligo_designer_toolsuite/oligo_specificity_filter/_filter_blastn.py
```diff
@@ -0,0 +1,189 @@
...
+        blast_matches_filtered = blast_matches.loc[
+            blast_matches.alignment_length > blast_matches.min_alignment_length
+        ]
+
+        probes_with_match = blast_matches_filtered["query"].unique()
+
+        return probes_with_match
+
+    def _filter_matching_probes(self, gene_DB, matching_probes):
+        """Filer out form the database the sequences with a match.
```
IsraMekki0, line 176: :return: gene_DB **without** the matching probes

oligo_designer_toolsuite/oligo_specificity_filter/_filter_bowtie.py
```diff
@@ -0,0 +1,209 @@
...
+from subprocess import Popen
+
+import pandas as pd
+from joblib import Parallel, delayed
+
+from . import SpecificityFilterBase
+
+
+class Bowtie(SpecificityFilterBase):
+    """This class filters probes based on the Bowtie short read alignment tool.
```
IsraMekki0, line 13: threshhold --> threshold

oligo_designer_toolsuite/oligo_specificity_filter/_filter_bowtie.py
```diff
@@ -0,0 +1,209 @@
...
+        super().__init__(dir_specificity)
+
+        if num_mismatches > 3:
+            raise ValueError(
+                "Choice of min_mismatches out of range for bowtie allignment tool. Please choose a value no greater than 3"
+            )
+        else:
+            self.num_mismatches = num_mismatches
+
+        if mismatch_region < 5:
```
IsraMekki0, line 39: What if mismatch region is None?

oligo_designer_toolsuite/oligo_specificity_filter/_filter_bowtie.py
```diff
@@ -0,0 +1,209 @@
...
+        )
+
+        # reconstruct the oligos_DB
+        for gene, filtered_oligo_DB in zip(genes, filtered_oligo_DBs):
+            oligo_DB[gene] = filtered_oligo_DB
+
+        return oligo_DB
+
+    def _run_bowtie(self, gene_DB, gene, index_name):
+        """Run Bowtie alignment tool to find regions of local similarity between sequences, where sequences are probes and transcripts.
```
IsraMekki0, line 92: Bowtie identifies all **alignments** between...
:param gene: id of **the** gene processed

oligo_designer_toolsuite/oligo_specificity_filter/_filter_bowtie.py
```diff
@@ -0,0 +1,209 @@
...
+        return probes_with_match
+
+    def _filter_matching_probes(self, gene_DB, matching_probes):
+        """Filer out form the database the sequences with a match.
+
+        :param gene_DB: dictionary with all the probes belonging to the current gene
+        :type gene_DB: dict
+        :param matching_probes: list of the probes with a match
+        :type matching_probes: list
+        :return: gene_DB withou the matching probes
```
IsraMekki0, line 201: :return: gene_DB **without** the matching probes

oligo_designer_toolsuite/oligo_specificity_filter/_filter_bowtie2.py
```diff
@@ -0,0 +1,203 @@
...
+from subprocess import Popen
+
+import pandas as pd
+from joblib import Parallel, delayed
+
+from . import SpecificityFilterBase
+
+
+class Bowtie2(SpecificityFilterBase):
+    """This class filters probes based on the Bowtie 2 read alignment tool. It is recommended to use Bowtie 2 instead of Bowtie for reads longer than about 50 bp, as it gives better performance.
```
IsraMekki0, line 13: The Bowtie 2 filter gives **an alignment** score to each ... the more **similar**...
 an **alignment** score greater than min_score...
    :param min_score: User defined **threshold** for **alignment** score. If specified, the Bowtie 2 filter filters out probes with an **alignment** score...

oligo_designer_toolsuite/oligo_specificity_filter/_filter_seed_region.py
```diff
@@ -0,0 +1,196 @@
...
+
+        # read the results of the bowtie search
+        bowtie_results = self._read_bowtie_output(file_bowtie_gene)
+        # filter the DB based on the bowtie results
+        matching_probes = self._find_matching_probes(bowtie_results)
+        filtered_gene_DB = self._filter_matching_probes(gene_DB, matching_probes)
+        # remove the temporary files
+        os.remove(os.path.join(self.dir_seed_region, file_bowtie_gene))
+        os.remove(os.path.join(self.dir_fasta, file_probe_fasta_gene))
+        return filtered_gene_DB
```
IsraMekki0, line 118: This part is shared with almost all other filters, could we refactor and put it in the base class?

oligo_designer_toolsuite/oligo_specificity_filter/_filter_seed_region.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def _read_bowtie_output(self, file_bowtie_gene):
```
IsraMekki0, line 120: Is this the same as _read_bowtie_output in the Bowtie filter? If so, create a base class with a common function

oligo_designer_toolsuite/oligo_specificity_filter/_filter_seed_region.py
```diff
@@ -0,0 +1,196 @@
...
+        :param blast_results: DataFrame with processed bowtie alignment search results.
+        :type blast_results: pandas.DataFrame
+        """
+        bowtie_matches = bowtie_results[
+            bowtie_results["query_gene_id"] != bowtie_results["reference_gene_id"]
+        ]
+        probes_with_match = bowtie_matches["query"].unique()
+        return probes_with_match
+
+    def _filter_matching_probes(self, gene_DB, matching_probes):
```
IsraMekki0, line 174: This should also go in the base class no?

oligo_designer_toolsuite/oligo_specificity_filter/_filter_seed_region.py
```diff
@@ -0,0 +1,196 @@
...
+        )
+        # return the real matches, that is the ones not belonging to the same gene of the query probe
+        bowtie_results["query_gene_id"] = bowtie_results["query"].str.split("_").str[0]
+        bowtie_results["reference_gene_id"] = (
+            bowtie_results["reference"].str.split("::").str[0]
+        )
+
+        return bowtie_results
+
+    def _find_matching_probes(self, bowtie_results):
```
IsraMekki0, line 161: Can also be refactored

oligo_designer_toolsuite/oligo_specificity_filter/_seed_region_creation.py
```diff
@@ -0,0 +1,129 @@
+from abc import ABC, abstractmethod
+
+
+class SeedRegionCreationBase(ABC):
+    """Base class tfor all the seed region creation classes"""
```
IsraMekki0, line 5: Base class **for** all the seed region creation classes

oligo_designer_toolsuite/oligo_specificity_filter/_seed_region_creation.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def __init__(self):
+        pass
+
+    @abstractmethod
+    def apply(self, oligo_DB):
+        """To each probe in the oligo DB add the  `start` and `end` values with key "seed_region_start" and "seed_region_end" which indicate the relative coordinates of the start and the end (included) of the seed region.
+
+        :param oligo_DB: database containin all teh oligo sequences and their features
```
IsraMekki0, line 14: database **containing** all **the** oligo sequences and their features

oligo_designer_toolsuite/oligo_specificity_filter/_seed_region_creation.py
```diff
@@ -0,0 +1,129 @@
...
+        :type oligo_DB: dict
+        """
+        return oligo_DB
+
+
+class SeedRegionCreationStandard(SeedRegionCreationBase):
+    """Standard seed region creation where the relative coordinates of the seed region are directly passed as arguments. If the end positionis bigger than teh length of the probe the seed region ends where teh probe ends.
```
IsraMekki0, line 21: ...**the** length of the probe the seed region ends where **the** probe ends.

oligo_designer_toolsuite/oligo_specificity_filter/_seed_region_creation.py
```diff
@@ -0,0 +1,129 @@
...
+        :param start: relative coordinate of the start of the seed region as a pecentage wrt to probe lenght
+        :type start: float
+        :param end: relative coordinate of the end of the seed region as a pecentage wrt to probe lenght
+        :type end: float
+    """
+
+    def __init__(self, start, end):
+        super().__init__()
+        if start < 0 or start > 1:
+            raise ValueError("Etart position must be in the interval [0,1]!")
```
IsraMekki0, line 64: **Start** position

oligo_designer_toolsuite/oligo_specificity_filter/_seed_region_creation.py
```diff
@@ -0,0 +1,129 @@
...
+                    round(self.start * length)
+                )
+                oligo_DB[gene][probe_id]["seed_region_end"] = int(
+                    round(self.end * length)
+                )
+        return oligo_DB
+
+
+class LigationRegionCreation(SeedRegionCreationBase):
+    """Seed regionused in the Paddlock probe designer pipeline, the region extends by `ligation_region_size` number of bases in both directions starting from the ligation site.
```
IsraMekki0, line 89: Seed **region used** in the **Padlock**...

oligo_designer_toolsuite/oligo_specificity_filter/_seed_region_creation.py
```diff
@@ -0,0 +1,129 @@
...
+                ligation_site = oligo_DB[gene][probe_id]["ligation_site"]
+                oligo_DB[gene][probe_id]["seed_region_start"] = int(
+                    max(ligation_site - (self.ligation_region_size - 1), 0)
+                )
+                oligo_DB[gene][probe_id]["seed_region_end"] = int(
+                    min(
+                        ligation_site + self.ligation_region_size,
+                        oligo_DB[gene][probe_id]["length"],
+                    )
+                )
```
IsraMekki0, line 128: This part can be refactored to the base class (with params)

oligo_designer_toolsuite/oligo_specificity_filter/_specificity_filters.py
```diff
@@ -0,0 +1,35 @@
+class SpecificityFilter:
+    """Thus class applies the specificity filters given in input sequentally to a ``CustomDB`` class.
```
IsraMekki0, line 2: **This** class applies the specificity filters given in input **sequentially** to a ``CustomDB`` class.

oligo_designer_toolsuite/oligo_specificity_filter/_specificity_filters.py
```diff
@@ -0,0 +1,35 @@
...
+
+    :param filters: List of all the filter classes that we want to apply to the database
+    :type filters: list of ``SpecificityFilterBase`` class
+    """
+
+    def __init__(self, filters):
+        self.filters = filters
+
+    def apply(self, DB, n_jobs=None, write_empty_genes=True):
+        """Applies all the class filter sequentailly to the ``oligos_DB`` sotored in the DB class given in input. Each filter parallelized and ``n_jobs`` represents the maximum number of cores available.
```
IsraMekki0, line 12: Applies all the class filter **sequentially** to the ``oligos_DB`` **stored** in the DB class...
:param n_jobs: maximum number of available jobs, if ``None`` **the** number of jobs of the ``DB`` class is used, defaults to None

oligo_designer_toolsuite/oligo_specificity_filter/_specificity_filters.py
```diff
@@ -0,0 +1,35 @@
...
+    """Thus class applies the specificity filters given in input sequentally to a ``CustomDB`` class.
+
+    :param filters: List of all the filter classes that we want to apply to the database
+    :type filters: list of ``SpecificityFilterBase`` class
+    """
+
+    def __init__(self, filters):
+        self.filters = filters
+
+    def apply(self, DB, n_jobs=None, write_empty_genes=True):
```
IsraMekki0, line 11: data_base instead of DB

oligo_designer_toolsuite/oligo_transcript_generation/_gene_transcript.py
```diff
@@ -23,17 +24,19 @@ def __init__(self, file_sequence, file_annotation):
...
+    ):
         """Creates a fasta file containing the whole transcriptome and its annotation file for the reference DB. The file contains also the exon junctions, which are the union of two consecuteive exons and
```
IsraMekki0, line 30: ...union of two **consecutive** exons...
        :param block_size: size of the exon **junctions**

oligo_designer_toolsuite/oligo_transcript_generation/_oligos.py
```diff
@@ -25,69 +27,68 @@ def __init__(self, probe_length_min, probe_length_max, file_sequence, filters):
...
+        :type n_jobs: int
         :param dir_output: directory where temporary diles are written, defaults to './output/annotation'
         :type dir_output: str, optional
         :return: the oligos DB containing all the probes created
         :rtype: dict
         """
 
-        def _get_probes(batch_id, genes_batch):
+        def _get_probes(gene, region_annotation):
             """Get the fasta sequence of all possible probes for all genes in the batch.
```
IsraMekki0, line 47: We do it genewise now, update the docstring

IsraMekki0 (commented):
Some typos and a few things to discuss later. Otherwise, it's fine by me

IsraMekki0 (approved):
Looks good to me :)

PR #23, opened 2023-01-19
Tutorial fix

francesco-campi:
Tutorial has 3 different config files, one for each source. Moreover, the custom fasta and gtf files are in the repository.


lisa-sousa (approved):

lisa-sousa:
Looks all good

PR #24, opened 2023-02-09
Refactoring

lisa-sousa:
The following modules were refactored:

- utils: 
  - exchanged GTF parser code to simpler code with fewer dependencies and able to read GTF and GFF3 files
  - moved out GFF parser to separate class
  - FTP download is now possible for all species via ensemble and NCBI
  - fasta header parser added
- database:
  - moved transcriptome creation to own more general class: genomic region generator
  - genomic region generator can now also generate CDS
  - fixed bed to gft conversion (bed files are 0-index based, while gtf files are 1-index based)
  - output of genomic region generator is fasta file
  - reference and oligo databases now only need fasta file as inputs
  - removed gene-specific dependencies in databases (e.g. transcript_id or exon_id)
  - fixed bug for split regions (e.g. exon junctions), where wrong coordinates were returned 
  - fixed coordinate calculation for minus strand where sequence is reverse complement
  - merge oligos first via coordinates (e.g. exon junction and exons lead to same oligo -> keep only exon information), then via sequence
  - changed unique ID of oligos from <region_name>_<number> to <region_name>_<chr>:<start>-<end>(<strand>)
- property filters
  - adjusted to new database structure
  - changed depratced GC function to GC_content function
  - added melting temp corrections (salt and chem correction) as optional
 
Furthermore, tests for those three modules were updated. They use now pytest instead of unit test. To keep temporary files, they use tmpdir_factory fixture.

Also, the tutorial was updated up to property filters.


oligo_designer_toolsuite/database/_genomic_region_generator.py
```diff
@@ -0,0 +1,702 @@
...
+
+
+############################################
+# Genomic Region Generator Classes
+############################################
+
+
+class CustomGenomicRegionGenerator:
+    """Class to generate sequences from annotated regions in GTF format and genomic fasta file.
+    Sequences are safed as fasta file with region id, additional information and coordinates in header.
```
francesco-campi, line 27: Typo: saved

oligo_designer_toolsuite/database/_oligos_database.py
```diff
@@ -1,444 +1,465 @@
...
 
-class CustomOligoDB:
-    """This class generates all possible guides that can be designed for a given list of genes,
-    based on the transcriptome annotation or the gene CDS or the whole genome.
-    Moreover, it can write to a file and read the database containing all the oligos sequences.
-    The gtf and a fasta files are passed in input.
+class OligoDatabase:
+    """This class generates all possible oligos that can be designed for a given list of regions (e.g. genes),
+    based on the transcriptome or the gene CDS annotation or the whole genome provided as fasta file.
+    Moreover, the database can be saved and loaded to/from a tsv file.
```
francesco-campi, line 31: Did we remove the option to write a gtf file?

lisa-sousa: yes, I removed the option for the gtf file, but added an option to just write the sequences with IDs to a fasta file. I think that's sufficient in the end. The tsv reader/writer is now the save/loader for the database

oligo_designer_toolsuite/database/_oligos_database.py
```diff
@@ -1,444 +1,465 @@
...
 
-    Species, genome_assembly, annotation_release are set to 'unknown' if thay are not given in input.
+    Source, species, annotation_release and genome_assembly are set to 'unknown' if thay are not given as input.
```
francesco-campi, line 33: Maybe we could add also here the specification that the fasta file has to meet

oligo_designer_toolsuite/database/_oligos_database.py
```diff
@@ -1,444 +1,465 @@
...
-        :param file_oligos_DB_fasta: Path to the file.
-        :type file_oligos_DB_fasta: str
-        :param file_oligos_DB_tsv: path of the oligos_DB file
-        :type file_oligos_DB_tsv: str
+        Creates the database containing all the oligo sequence extracted form the given sequences (fasta file)
+        and belonging the the specified region. If no list of region IDs is specified then all regions available
+        in the fasta file will be used. The database created is not written automatically to the disk,
+        the ``save_oligo_database`` method has to be called separately.
 
+        :param region_ids: List of regions for which the oligos should be generated, defaults to None
```
francesco-campi, line 137: Can the region ID be different form the gene id? Otherwise it seems to unclear how this list should look like

lisa-sousa: yes, it should just be an id that is used in the fasta header, e.g. if someone uses enhancer regions instead of genes, they could give those regions IDs as well (e.g. enhancer_xyz) and a subset can be given as list of region ids to filter on. It just makes it more generic to other regions (not restricted to genes).

oligo_designer_toolsuite/database/_reference_database.py
```diff
@@ -1,276 +1,130 @@
...
-            annotation_file,
-            sequence_file,
-            species,
-            genome_assembly,
-            annotation_release,
-            files_source,
-            dir_output,
-        )
+        database_filtered = []
+        for entry in self.database:
```
francesco-campi, line 126: here the database has to be filled beforehand, amd in case it is empty it does not give any warning. we could directly call the `load_fasta_into_database` function here.

francesco-campi (approved):
Lokks good to me!

francesco-campi (approved):

francesco-campi (approved):

PR #34, opened 2023-12-08
Update specificity

IsraMekki0:
1. Replaced hardcoded separators with constants from `oligo_designer_toolsuite.constants`.
2. Refactored specific filtering (e.g. `_run_bowtie`) into `_create_index`, `_run_search` and `_run_filter`.
3. Updated private functions to take the entire database dict as input instead of `database_region`.
4. TODO: Replace dict database with `OligoDatabase` and remove `_create_fasta_file` and  `_create_fasta_multiple_regions` in favor of `OligoDatabase.write_fasta_from_database`


francesco-campi (commented):
A few comments

PR #38, opened 2024-02-09
Feature/refactor_property_filter

francesco-campi:
Sorry for the confusion, during the last update meeting I suggested that the filters `HomopolymericRunsFilter` and `ProhibitedSequenceFilter` shoud be able to look for multiple sequences/homopolimericruns at the same time. 

Therefore, now both functions can recieve in input coth a single seqeunce or a list of seqeunces. Let me know if you find it clear especially from the documentation.


IsraMekki0 (approved):
Looks good to me. I would just avoid testing filters in the database test

PR #55, opened 2024-03-26
49 exact match filter

lisa-sousa:
bug fixed of exact match filter logic:
- for get_oligo_pair_hit we compare sequence to their reverse complements instead of the same sequence, e.g. oligo to target and target to oligo
- for apply we compare sequence to same sequence, e.g. oligo to oligo and target to target

in addition, we added an optional filter policy for the apply function such that hits only get removed from one oligoid and not both. Which oligo id to remove is determied by the policy.

Last, we fixed some bug in the Blast filter stemming from the introduction of the OligoAttributes class.


oligo_designer_toolsuite/oligo_specificity_filter/_filter_exact_matches.py
```diff
@@ -45,35 +50,45 @@ def apply(self, sequence_type: _TYPES_SEQ, oligo_database: OligoDatabase, n_jobs
...
                 search_results=search_results,
                 sequence_oligoids_mapping=sequence_oligoids_mapping,
+                consider_hits_from_input_region=False,
             )
             for region_id in region_ids
         )
 
-        # Process results
-        table_hits, oligos_with_hits = zip(*results)
+        if self.policy:
```
IsraMekki0, line 75: I think we should have a dedicated policy that simply deletes both oligos, which would be the default.

oligo_designer_toolsuite/oligo_specificity_filter/_filter_exact_matches.py
```diff
@@ -45,35 +50,45 @@ def apply(self, sequence_type: _TYPES_SEQ, oligo_database: OligoDatabase, n_jobs
...
+                graph=nx.from_edgelist(oligo_pair_hits), oligo_database=oligo_database
+            )
+        else:
+            oligos_with_hits = {}
+            for region_id, table_hits_region in zip(region_ids, table_hits):
+                oligos_with_hits[str(region_id)] = table_hits_region["query"].unique()
 
-        for region_id, oligos_with_hits_region in zip(region_ids, oligos_with_hits):
+        for region in region_ids:
             database_region_filtered = self._filter_hits_from_database(
```
IsraMekki0, line 87: We can run this in parallel

oligo_designer_toolsuite/oligo_specificity_filter/_filter_exact_matches.py
```diff
@@ -101,35 +116,38 @@ def get_oligo_pair_hits(
         assert (
             sequence_type in options
         ), f"Sequence type not supported! '{sequence_type}' is not in {options}."
+        sequence_type_reverse_complement = options[0] if options[0] != sequence_type else options[1]
 
         # extract all the sequences
-        search_results = self._get_duplicated_sequences(sequence_type, oligo_database)
+        sequences = oligo_database.get_sequence_list(sequence_type=sequence_type)
+        sequences_rc = oligo_database.get_sequence_list(sequence_type=sequence_type_reverse_complement)
+        search_results = self._get_duplicated_sequences(sequences + sequences_rc)
```
IsraMekki0, line 124: I think we should give the user a choice whether to filter duplicated sequences, sequences that are reverse complement to each other, or both (param in the `__init__`)

IsraMekki0 (changes requested):
Two main points:

- Create a new policy to remove both oligos
- Make it possible to chose the mode of the exact match filter (duplicates, complementary, both)

oligo_designer_toolsuite/oligo_specificity_filter/_filter_exact_matches.py
lisa-sousa: The rationale behind this is that we have two application cases:
1. apply function -> using exact matches as filter: here we want to remove oligos which are duplicated in the database, i.e. we use the same forward sequences
2. oligo_pair_hits function -> using exact matches to get hits table for cross-hybridization filter. Here we have to make sure to find the exact matches among the reverse complements, hence, we match all forward sequences to all reverse complements

oligo_designer_toolsuite/oligo_specificity_filter/_filter_exact_matches.py
lisa-sousa: we actually never used parallel loops in this case because this step is not compute intensive and might cost more to setup the threads in this case

PR #56, opened 2024-03-28
Changed instances of dictionaries to LRUDDBDict

IsraMekki0:


lisa-sousa (commented):
Looks good. I just changed the part where you always setup a cache path for a dict and instead stored the cache path as class property.

IsraMekki0:
I think the cache path should be as it was because it's supposed to be deleted anyway, I don't want to user to even see that there are files.

PR #65, opened 2024-05-29
Implemented heterodimer filter

IsraMekki0:
- [x] Created Heterodimer filter
- [x] Added tests


oligo_designer_toolsuite/oligo_property_filter/__init__.py
```diff
@@ -49,6 +51,7 @@
     "GCClampFilter",
     "MeltingTemperatureNNFilter",
     "HomodimerFilter",
```
lisa-sousa, line 53: We should rename both "Homodimer" and "heterodimer" filters because we are not really checking for the thermodynamics of the sequence binding as they do here: https://libnano.github.io/primer3-py/api/thermoanalysis.html
Maybe rename to self_complement and complement filter or similar

lisa-sousa (changes requested):
I just left one comment regarding the naming of the filters, which we should change.

PR #67, opened 2024-06-18
Enhance database loading

lisa-sousa:
Adjusted the code for big data loading:
- use the database structure correctly, i.e. using threads in parallel loops to not copy the database for each parallel process which created a memory overload but instead modify database in place
- bug fix the access of the database for parallel processing, i.e. make sure to only access one region per proccess to not have data issues in parallel processes due to simultaneous access of database
- bug fix region removal from database, which did not work properly due to bug in effidict -> currently using workaround until fixed in effidict
- adjust tests
- adjust pipelines


IsraMekki0 (approved):
Looks good to me

PR #68, opened 2024-07-02
remove Bio package for blast and turn into subprocess

lisa-sousa:
Removed all usage of Bio.Blast.Applications and call all alignment tools via subproccess now.


IsraMekki0 (approved):
Looks good to me

PR #78, opened 2024-08-02
62 seqfish pipeline

lisa-sousa:
Created seq-fish pipeline with 

- primer generation
- readout probes generation
- codebook implementation
- all required property filters
- correct blast specificity parameters
- correct blast corss-hybridization parameters
- correct set scoring parameters 
- distance parameter between probes


francesco-campi (approved):
All good!

PR #86, opened 2024-08-13
Faster heuristic method to generate non-overlapping sets

IsraMekki0:
Issue: when generating bigger sets the initial algorithm becomes too slow

### DIFF
--- a/oligo_designer_toolsuite/oligo_selection/_generate_oligosets.py
+++ b/oligo_designer_toolsuite/oligo_selection/_generate_oligosets.py
@@ -252,19 +252,66 @@ class OligosetGeneratorIndependentSet:
         :return: A DataFrame containing the best non-overlapping oligo sets.
         :rtype: pd.DataFrame
         """
+        n_attempts = 1000
+
+        def _find_cliques_of_size_n_and_more(G, n, max_attempts=n_attempts):
+            cliques = set()  # Using a set to avoid duplicate cliques
+            attempts = 0
+            big_enough_clique_found = False
+            prev_max_clique = None
+
+            while attempts < max_attempts and not big_enough_clique_found:
+                max_clique = nx.approximation.max_clique(G)
+
+                if max_clique == prev_max_clique:
+                    break
+
+                if len(max_clique) >= n:
+                    sub_cliques = list(nx.find_cliques(G.subgraph(max_clique)))
+                    for clique in sub_cliques:
+                        if len(clique) >= n:
+                            big_enough_clique_found = True
+                            cliques.add(tuple(sorted(clique)))  # Sort and convert to tuple for set uniqueness
+
+                prev_max_clique = max_clique
+                attempts += 1
+
+            return [list(clique) for clique in cliques]
+
+        def bron_kerbosch_with_pivot(G, R, P, X, n):
+            if len(R) == n:
+                return R
+            if not P and not X:
+                return []
+            for v in list(P):
+                new_R = R.union([v])
+                new_P = P.intersection(G.neighbors(v))
+                new_X = X.intersection(G.neighbors(v))
+                result = bron_kerbosch_with_pivot(G, new_R, new_P, new_X, n)
+                if result:
+                    return result
+                P.remove(v)
+                X.add(v)
+            return []
+
+        def find_clique_of_size_n(G, n):
+            P = set(G.nodes())
+            return bron_kerbosch_with_pivot(G, set(), P, set(), n)
+
         # Represent overlap matrix as graph
         G = nx.convert_matrix.from_numpy_array(overlapping_matrix.values)
         G = nx.relabel_nodes(G, {i: overlapping_matrix.index[i] for i in range(len(oligos_scores.index))})
 
         # First check if there are no cliques with n oligos
-        cliques = nx.algorithms.clique.find_cliques(G)
         n = self.opt_oligoset_size
-        n_max = 0
 
-        for clique in cliques:
-            n_max = max(len(clique), n_max)
-            if n_max >= n:
-                break
+        biggest_clique = find_clique_of_size_n(G, n)
+        n_max = len(biggest_clique)
+        # for _ in range(n_attempts):
+        #     biggest_clique = nx.approximation.max_clique(G)
+        #     n_max = len(biggest_clique)
+        #     if n_max >= n:
+        #         break
 
         if n_max < n:
             if n_max <= self.min_oligoset_size:  # in this case we don't need to compute the sets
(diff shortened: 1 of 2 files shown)


oligo_designer_toolsuite/pipelines/_seqfish_plus_probe_designer.py
```diff
@@ -267,19 +267,30 @@ def create_oligo_sets(
             GC_content_opt=GC_content_opt, GC_weight=GC_weight, UTR_weight=UTR_weight
         )
         set_scoring = LowestSetScoring(ascending=True)
+
+        selection_policy = GreedySelectionPolicy(
```
lisa-sousa, line 271: compare to graph based with pre-filter

oligo_designer_toolsuite/pipelines/_oligo_seq_probe_designer.py
```diff
@@ -450,19 +448,27 @@ def create_oligo_sets(
...
-            heuristic_selection=heuristic_selection_independent_set,
+            heuristic_selection=selection_policy,
             max_oligos=max_oligos,
             distance_between_oligos=distance_between_oligos,
         )
         oligo_database = oligoset_generator.apply(
             oligo_database=oligo_database,
             sequence_type="oligo",
-            n_sets=n_sets,
+            pre_filter=False,
```
lisa-sousa, line 471: move to yaml file

oligo_designer_toolsuite/oligo_selection/_generate_oligosets.py
```diff
@@ -144,20 +149,21 @@ def _get_oligo_set_for_gene(
         oligos_scores.sort_values(ascending=self.ascending, inplace=True)
 
         # hard limit on the number of oligos
-        if len(oligos_scores) > self.max_oligos:
-            # select the best oligos
-            for oligo_id in oligos_scores.index[self.max_oligos :]:
-                del oligo_database.database[region_id][oligo_id]
-                oligos_scores.drop(oligo_id, inplace=True)
+        # if len(oligos_scores) > self.max_oligos:
```
lisa-sousa, line 152: move to graph based

IsraMekki0: It would have been too (unnecessarily) complicated to move it to the graph-based class. Adding the isinstance check for now

IsraMekki0: Initialize max_oligos to None (infinity) by default

oligo_designer_toolsuite/oligo_selection/_heuristic_selection_methods.py
```diff
@@ -2,79 +2,392 @@
...
-    :rtype: tuple(dict, pd.Series, pd.Series)
+    :param set_scoring: Set scoring class.
+    :type set_scoring: SetScoringBase
     """
+    def __init__(
+        self, set_size_opt: int, set_size_min: int, n_sets: int, ascending: bool, set_scoring: SetScoringBase
+    ) -> None:
+        self.opt_oligoset_size = set_size_opt
+        self.min_oligoset_size = set_size_min
+        self.n_sets = n_sets
```
lisa-sousa, line 40: this should be parametrized from the"OligosetGeneratorIndependentSet" class -> you have those parameters in the init but you don't use them in the class.

oligo_designer_toolsuite/oligo_selection/_generate_oligosets.py
```diff
@@ -52,7 +53,7 @@ def __init__(
         min_oligoset_size: int,
         oligos_scoring: OligoScoringBase,
         set_scoring: SetScoringBase,
-        heuristic_selection: Callable = None,
+        heuristic_selection: OligoSelectionPolicy,
         distance_between_oligos: int = 0,
         max_oligos: int = 5000,
```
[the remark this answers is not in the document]
IsraMekki0: Hmm, good point. The logic is to be conservative: we limit so that in the worst case (GraphBased, which is the default approach) doesn't crash.. What do you think?

oligo_designer_toolsuite/oligo_selection/_generate_oligosets.py
```diff
@@ -71,7 +72,8 @@ def apply(
         self,
         oligo_database: OligoDatabase,
         sequence_type: _TYPES_SEQ,
-        n_sets: int = 50,
+        pre_filter: bool,
```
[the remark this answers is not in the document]
IsraMekki0: I see it more like an additional step. The policy is more like for how we select, and the pre-filtering is for the initial set basically. It would also make sense to put it in the init though. No strong preferences for me. Would you prefer in the Policy instead?

oligo_designer_toolsuite/oligo_selection/_heuristic_selection_methods.py
```diff
@@ -2,79 +2,392 @@
...
+    :param ascending: Flag indicating whether to sort scores in ascending order (depends on the score, set to True if lower scores are better).
     :type ascending: bool
-    :param n_trials: Number of top scoring oligos to consider for set formation, defaults to 100.
-    :type n_trials: int
-    :return: Updated scores, and the best set found.
-    :rtype: tuple(dict, pd.Series, pd.Series)
+    :param set_scoring: Set scoring class.
+    :type set_scoring: SetScoringBase
     """
+    def __init__(
```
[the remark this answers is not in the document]
IsraMekki0: (depends on what we decide above)

oligo_designer_toolsuite/oligo_selection/_heuristic_selection_methods.py
```diff
@@ -2,79 +2,392 @@
...
+        :param n_attempts: Number of attempts to generate valid sets.
+        :type n_attempts: int
+        :return: DataFrame containing selected oligo sets with scores.
+        :rtype: pd.DataFrame
+        """
+
+        if pre_filter:
+            oligos_scores, overlapping_matrix = self.pre_filter_oligos(oligos_scores, overlapping_matrix)
+
+        if len(oligos_scores) > 0:
```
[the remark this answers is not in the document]
IsraMekki0: Good point 👍

oligo_designer_toolsuite/oligo_selection/_heuristic_selection_methods.py
```diff
@@ -2,79 +2,392 @@
...
+                    available_oligos = adjusted_scores.sort_values(ascending=self.ascending).index.to_list()
+
+            if len(selected_sets) == self.n_sets:
                 break
+        selected_sets = [list(selected_set) for selected_set in selected_sets]
+
+        return selected_sets
+
+
+class GraphBasedSelectionPolicy(OligoSelectionPolicy):
```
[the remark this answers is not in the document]
IsraMekki0: All we need is in the parent class

oligo_designer_toolsuite/oligo_selection/_heuristic_selection_methods.py
IsraMekki0: I agree there's a problem, but I'd say actually remove them from `OligosetGeneratorIndependentSet` because they are not used there. Otherwise, I'd have to supply them through the apply function, no?

oligo_designer_toolsuite/oligo_selection/_generate_oligosets.py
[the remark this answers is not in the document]
lisa-sousa: I would prefer in the heuristic

oligo_designer_toolsuite/oligo_selection/_heuristic_selection_methods.py
lisa-sousa: just keep thge parameters in the selection policy

oligo_designer_toolsuite/oligo_selection/_heuristic_selection_methods.py
```diff
@@ -2,79 +2,392 @@
...
-def heuristic_selection_independent_set(
-    oligoset_init: list,
-    oligos_scores: pd.Series,
-    overlapping_matrix: csr_matrix,
-    overlapping_matrix_ids: list,
-    n_oligo: int,
-    ascending: bool,
-    n_trials: int = 100,
-):
+class OligoSelectionPolicy:
```
lisa-sousa, line 20: rename heuristic to selection ploicies

lisa-sousa (approved):
All comments have been addressed. Good to merge

PR #92, opened 2024-12-18
89 refactor pipelines

lisa-sousa:
- Refactored each pipeline ot have one main class that merges all subparts of the pipeline, e.g. target, readout probe and primer generation.
- Updated all doc string for all pipelines
- Updated tests for pipelines


IsraMekki0 (approved):
Looks good 😊

PR #113, opened 2025-03-18
112 general implement missing tests

IsraMekki0:
Added tests for HomogeneousPropertyOligoSetGenerator and unit tests for OligoSelectionPolicy (Greedy and graph based)
Some notes:
- Doesn't include integration test of GreedyOligoSelectionPolicy into OligosetGeneratorIndependentSet because the tests already include the integration of GraphBasedOligoSelectionPolicy. I don't think it's necessary to test both because GreedyOligoSelectionPolicy is strictly defined to have the same interface as GraphBasedOligoSelectionPolicy and testing the integration of one of them is enough.
- `unused_homogeneous_oligosets.tsv` is unused 😛 but it's a sample that could be useful. I had created it to test HomogeneousPropertyOligoSetGenerator initially, but it turns out the results are not stable. You can delete it if you don't think we should keep it


lisa-sousa (approved):
Looks good

PR #121, opened 2025-06-27
120 fix intergenic parsing

### LINKED ISSUE #120: Intergenic regions can have ambiguous region_id
When using `genomic_region_generator` on the latest NCBI homo sapiens annotation, the created intergenic regions are not unique. The `region_id` can be contained twice. One example is InterRegMinus10_100, which is associated with located on chromosome 10 with coordinates 120981654-120984961 and another with coordinates 3172842-3232263. Both are on the minus strand.

This was tested on the 115-cyclehcr-pipeline-create-base-pipeline branch and occurred both when parsing genes and intergenic regions or only intergenic regions.

jonas-hag:
Closes #120 

Generating intergenic regions led to duplicated regions. This was due to 2 different bugs:
- in `parse_annotation_from_gff`, pandas reads in a csv file with information about the sequences, containing the chromosome. Because the file is so large, `pd.read_csv` reads in the file in chunks, which leads to guessing the dtype. Because besides the autosomes, also scaffold chromsomes with alphanumeric names are contained, some chunks contain only autosomes and other chunks both autosomes and scaffold chromosomes. Therefore, the same chromosome can be read in as integer and as string, e.g. as 1 and '1'.  Because `_compute_intergenic_annotation` splits the dataframe based on unique values, the entries for the autosomes get split up into two parts and duplicates are generated.
This behavior of `read_csv` only occurs with large files and therefore was not caught by small test files.
- in `_compute_intergenic_annotation_strand`, in the case if one strand of a chromosome doesn't contain any genes, the intergenic region was always described as InterRegPlus. This mainly affected scaffold chromosomes.
Additionally, I changed the naming of the only intergenic region from _1 to _0 to be in line how the first intergenic region is called when there are several intergenic regions.

For the parsing, I added a test case with a large enough file that contains an autosome and a scaffold chromosome that originally showed the wrong parsing behavior.

The changes in how the imports and exceptions are written is due to the pre-commit hook.

### DIFF
--- a/oligo_designer_toolsuite/sequence_generator/_genomic_region_generator.py
+++ b/oligo_designer_toolsuite/sequence_generator/_genomic_region_generator.py
@@ -291,13 +291,19 @@ class CustomGenomicRegionGenerator:
                 chromosome_length = pd.read_csv(
                     file_chromosome_length, sep="\t", comment="t", header=0, names=["seqid", "length"]
                 )
+                if strand == "+":
+                    region_id_name = "InterRegPlus"
+                elif strand == "-":
+                    region_id_name = "InterRegMinus"
+                else:
+                    raise ValueError(f"Invalid strand value: {strand}. Expected '+' or '-'.")
                 intergenic_annotation = pd.DataFrame(
                     {
                         "seqid": seqid,
                         "start_0base": 0,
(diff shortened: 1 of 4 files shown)


lisa-sousa (approved):
looks good

PR #124, opened 2025-07-03
Add stride length option for sliding window

### LINKED ISSUE #122: Stride length for sliding window
When creating oligos in large scale (in this case to generate the data set for hybridization filter experiments), it would be great if the stride length of the sliding window could be changed.

jonas-hag:
- add the possibility to specify the stride of the sliding window when creating oligos
- added tests for this -> is there a better way to do it?
- fixed tests that check for the oligo length
- added tests for oligo generation via sliding window for exon-exon junctions

Please check especially if it still is ok for genes on the minus strand


lisa-sousa (approved):
Looks all good

Issue #125, opened 2025-07-14
Naming convention for reference data files


jonas-hag:
Create a structured naming convention for files that store the reference data (FASTA files):
- how to best separate the parameters in the file names?
- include the block size for exon-exon junctions in the file name

This is important for caching for the frontend

PR #126, opened 2025-07-21
naming of FASTA files with extracted regions

jonas-hag:
This is related to #125 as @yarkinerenn mentioned it's probably easier than hashing the parameters used.

change the naming of the FASTA files when the GenomicRegionGenerator is used to extract the regions based on the different annotations

- all parameter names and parameter values are separated by `__`
- the annotation type is moved to the end of the file name
- all parameters used in the `get_sequence` functions are included in the file names

The thing I'm a bit worried about is the file name length as on many systems the total path length is limited to 255 or 260 characters. @lisa-sousa what do you think?

### DIFF
--- a/oligo_designer_toolsuite/sequence_generator/_genomic_region_generator.py
+++ b/oligo_designer_toolsuite/sequence_generator/_genomic_region_generator.py
@@ -2,10 +2,10 @@
 # imports
 ############################################
 
-import os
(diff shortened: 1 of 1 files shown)


lisa-sousa (approved):
I removed the collapse_regions info since this is basically always done in our pipelines

PR #128, opened 2025-08-26
add warning if no oligos are created

### LINKED ISSUE #127: Scrinshot creates an empty fasta files per gene and it creates an unexpected error
While I running the scrinshot with different genomic regions I stumbled upon this error. oligo_designer_toolsuite/utils/_sequence_parser.py(299)check_fasta_format()
-> raise ValueError("Fasta file has incorrect format!" 

while using this particular genomic region set up.

Image

with the scrinshot parameters that are default on the config file of the project.

The error stems from this issue,

Image

pipeline creates this middle fasta files per gene, and when the fasta files are empty(in this case AARS1.fna) it throws the error.

jonas-hag:
add warning if `create_sequences_sliding_window` doesn't create any oligos for a specific region. this typically happens when the minimum sequence length is longer than the input sequence, e.g. for exon-exon junctions with small block sizes

this also fixed #127 because empty FASTA files get deleted after the warning and can't cause an "ValueError("Fasta file has incorrect format!" in `check_fasta_format`


lisa-sousa (approved):
Looks all good

PR #131, opened 2025-09-15
Unify blast and bowtie `search_parameters`

### LINKED ISSUE #130: Additional arguments for blast and bowtie
Currently, the additional arguments for bowtie in `search_parameters` needs to contain the leading `-`, whereas for blast the dash is automatically added. Unify this to how we handle the arguments for blast.

Also, currently for bowtie `-a` is fixed. Check the `search_parameters` for `k` and remove `a` if necessary.

jonas-hag:
EDIT: see below comments

unify the behavior of `search_parameters` for bowtie and blast filters. Now, as for blast, the arguments in `search_parameters` don't need to be specified with a leading `-`.

Also, we check if `k` is in `search_parameters` for the bowtie filters (specify the number of reported alignments) and in this case don't use the `-a` option anymore.

closes #130

### DIFF
--- a/oligo_designer_toolsuite/oligo_specificity_filter/_filter_blastn.py
+++ b/oligo_designer_toolsuite/oligo_specificity_filter/_filter_blastn.py
@@ -3,13 +3,13 @@
 ############################################
 
 import os
-import warnings
 import subprocess
+import warnings
+from abc import abstractmethod
+from typing import List, Tuple, Union
+
 import numpy as np
 import pandas as pd
-
-from abc import abstractmethod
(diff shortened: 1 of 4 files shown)


jonas-hag:
Because bowtie arguments can have leading `-` or `--`, I've changed the implementation that now the format for blast arguments is the same as bowtie arguments, i.e., you have to explicitly specify the dash.

Our Blast filters also have `hit_parameters` which don't take leading dashed, which is a bit confusing, but they aren't passed to blastn, instead, they are used in our python code how to deal with the results. Therefore, I left them as they are.

PR #133, opened 2025-10-07
Unify blast and bowtie `search_parameters`

### LINKED ISSUE #130: Additional arguments for blast and bowtie

jonas-hag:
Unify the behavior of search_parameters for bowtie and blast filters. Because bowtie arguments can have leading - or --, I've changed the implementation that the format for blast arguments is the same as bowtie arguments, i.e., you have to explicitly specify the dash.

Our Blast filters also have hit_parameters which don't take leading dashed, which is a bit confusing, but they aren't passed to blastn, instead, they are used in our python code how to deal with the results. Therefore, I left them as they are.

The tests and config files for pipelines were changed accordingly.

Additionally, now bowtie filters can use the `-k` option in the optional arguments passed to bowtie instead of the default `-a` option -> return only some and not all alignments

I tried to write a test for the `-k` vs. `-a` option, however, directly using `_run_search` from `BowtieFilter` doesn't work because there is no way to set `sequence_type`. Therefore, at the moment no test is included.


lisa-sousa (approved):
Looks good. Thanks

Issue #143, opened 2025-11-18
[General] Pipeline Configuration Refactoring with Pydantic


lisa-sousa:
## Overview

Refactor all pipeline configurations to use Pydantic models, moving defaults from `set_developer_parameters()` method signatures to centralized config classes. This eliminates redundancy, adds validation, and simplifies YAML config handling.

## Current Problems

- Default values duplicated in `set_developer_parameters()` method signatures
- Verbose main() functions manually mapping YAML dict to method parameters
- No validation of config values from YAML files
- No type safety or IDE autocomplete support
- Long parameter lists make methods hard to maintain

## Solution Approach

1. Create reusable Pydantic models for common parameter groups
2. Create pipeline-specific config models with all defaults
3. Simplify `set_developer_parameters()` to accept config object
4. Update main() functions to validate YAML and pass config object
5. Maintain backward compatibility with existing YAML configs

### Step 1: Add Pydantic Dependency

- Add `pydantic>=2.0` to `pyproject.toml` dependencies
- Verify compatibility with Python >=3.10

### Step 2: Create Shared Parameter Models

Create `oligo_designer_toolsuite/pipelines/_config_models.py` with reusable models:

- `BlastnSearchParameters`: Common BLASTN search parameters (perc_identity, strand, word_size, dust, soft_masking, max_target_seqs, max_hsps)
- `BlastnHitParameters`: Hit filtering parameters (coverage, min_alignment_length)
- `TmParameters`: Melting temperature parameters (nn_table, tmm_table, imm_table, de_table, dnac1, dnac2, saltcorr, Na, K, Tris, Mg, dNTPs, check, strict, c_seq, shift, selfcomp)
- `TmChemCorrectionParameters`: Chemical correction parameters (DMSO, fmd, DMSOfactor, fmdfactor, fmdmethod, GC)
- `TmSaltCorrectionParameters`: Salt correction parameters (optional)
- `OligoSetSelectionParameters`: Set selection parameters (max_graph_size, n_attempts, heuristic, heuristic_n_attempts)

Note: Handle dict keys with dashes (e.g., `-perc_identity`) by using Field aliases or custom validators.

### Step 3: Create Pipeline-Specific Config Models

For each pipeline, create a config model in `_config_models.py`:

1. **MerfishProbeDesignerConfig**: 

- User params: n_jobs, dir_output, write_intermediate_steps, top_n_sets, file_regions, etc.
[...]

PR #166, opened 2026-04-22
Mitigate code injection vulnerabilities via subprocess calls

### LINKED ISSUE #162: Proposal: Identify and mitigate code injection vulnerabilities
As the library wasn't originally intended to operate on untrusted user input, there are code injection vulnerabilities present. This is only problematic for scenarios with input provided by potentially malicious users (e.g. cloud). This shouldn't be the case for anyone yet, hence the public issue.

As an example, the current input parsing logic on the main branch allows a user to input a modified Scrinshot config:
```yaml
[...]
target_probe_cross_hybridization_blastn_search_parameters:
  perc_identity: 80 # the higher the percent identity is, the more significant the match
  strand: "minus" # this parameter is fixed
  outfmt: "6 qseqid sseqid length qstart qend qlen"
  word_size: "10\"; touch ../../RemoteCodeExecution.sh; exit;\""
[...]
```
The raw concatenation of command line arguments allows a malicious user to execute arbitrary shell code, in this case on a server running ODT.

This issue is intended as a tracking issue for finding and mitigating these vulnerabilities as there might be more than this one example. The ongoing work on integrating Pydantic might already solve most if not all of these problems.

maxtidev:
### Summary

This PR changes the implementation of subprocess calls such that commands are passed to `subprocess.run` as a list of arguments instead of a raw string. By removing `shell=True` we prevent running arbitrary shell commands as laid out in #162. See "Use shell=True with caution" in [this article](https://snyk.io/de/blog/command-injection-python-prevention-examples/) for more details.

Changes:

- call subprocesses without `shell=True` by passing list of arguments
- change usage of `subprocess.Popen` to [recommended](https://docs.python.org/3.12/library/subprocess.html#using-the-subprocess-module) `subprocess.run`
- remove custom quoting/escaping (this gets handled by `subprocess.run`)
- add `check=True` such that ODT will now raise if a subprocess fails
- add `-f` parameter to `bcftools index` in `merge_vcf_files()` as the subprocess could fail due to already existing files otherwise

I decided not to change the Jupyter notebooks that also use `subprocess.Popen` since they're not part of the library and seem to be used for local testing only. Some Python tests failed for me both on the `dev` branch as well as on this feature branch so I assume that is unrelated.

This PR will slightly conflict with #149. Since this vulnerability is currently blocking deployment of ODT Cloud and any external workarounds are more complex than this direct fix here, I'll gladly help integrate the changes if desired.

Closes #162 for now. We will create a new issue should we find another potential vulnerability.

### DIFF
--- a/oligo_designer_toolsuite/oligo_specificity_filter/_filter_blastn.py
+++ b/oligo_designer_toolsuite/oligo_specificity_filter/_filter_blastn.py
@@ -117,9 +117,11 @@ class BlastNFilter(AlignmentSpecificityFilter):
             filename=f"db_reference_{self.filter_name}",
             dir_output=self.dir_output,
         )
+
         ## Create blast index
-        cmd = "makeblastdb -dbtype nucl" + " -out " + file_reference + " -in " + file_reference
-        process = subprocess.Popen(cmd, shell=True, cwd=self.dir_output, stdout=subprocess.DEVNULL).wait()
+        args = ["makeblastdb", "-dbtype", "nucl", "-out", file_reference, "-in", file_reference]
+
+        subprocess.run(args, cwd=self.dir_output, check=True)
 
         return file_reference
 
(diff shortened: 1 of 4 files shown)


jonas-hag (approved):
thank you for the PR!

looks good to me; we have also noticed the `-f` issue and the current failing tests are addressed in #143

PR #170, opened 2026-05-05
Prevent unexpected path resolution using safe_append_filename helper

simon-ament:
### Summary

This PR introduces a `safe_append_filename` helper function to replace `os.path.join` calls where dynamically generated filenames are appended to a directory path for file placement or lookup. These filenames are often derived from user input, which can lead to unexpected behavior when they contain special characters or sequences like `/`, `/../`, or `\` (depending on OS).

The goal of this PR is to prevent the creation of unintended subdirectories or traversal outside the target directory by ensuring that the resolved path remains strictly within the specified directory.

### Path Traversal Mitigation

Since all existing `os.path.join` calls (outside of tests) operate on files in the `output_dir`, this PR also helps mitigate path traversal vulnerabilities when handling untrusted user input as long as the `output_dir` itself is trusted.

### What changed

1. `safe_append_filename` introduced in `_checkers_and_helpers.py`
- This function joins a directory path and a filename, then resolves the resulting path
- It raises an exception if the resolved path falls outside the specified directory or if the filename is altered during resolution
2. Replaced `os.path.join` calls where filenames are dynamically generated
3. Removed duplicate join with `output_dir` before blastn subprocess call, since `file_reference` already holds a path pointing into the `output_dir`

### DIFF
--- a/oligo_designer_toolsuite/database/_oligo_database.py
+++ b/oligo_designer_toolsuite/database/_oligo_database.py
@@ -32,6 +32,7 @@ from oligo_designer_toolsuite.utils import (
     format_oligo_properties,
     merge_databases,
 )
+from oligo_designer_toolsuite.utils._checkers_and_helpers import safe_append_filename
 
 CustomYamlDumper.add_representer(list, CustomYamlDumper.represent_list)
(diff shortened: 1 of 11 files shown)


jonas-hag (approved):
thank you, looks good!

PR #171, opened 2026-05-06
Make pipelines callable via function calls with Pydantic models

### LINKED ISSUE #151: Make pipelines callable as Python module
Currently, if you want to start a pipeline via a Python script, you can either call the installed script as a subprocess call like this:
```python
subprocess.run(
            ["scrinshot_probe_designer", "-c", config_path],
            capture_output=True,
            text=True
        )
```
...or you import the script directly, for example like this:
```python
from oligo_designer_toolsuite.pipelines._scrinshot_probe_designer import main as scrinshot_probe_designer

scrinshot_probe_designer() # no easy way of passing the config file argument
```
For the use case of running an entire pipeline via a Python script, both of these options seems suboptimal. The first option is unwieldy and decouples the pipeline from the rest of the Python script, hindering tighter integration (exception-catching, parameter type-checking, debugging). The other option just runs the script and expects the arguments to have been passed via the command line which isn't always what we want.

To make this nicer, I propose decoupling the cli arguments parsing from the main pipeline execution. For example, we could have a `pipelines._scrinshot_probe_designer_cli` that reads the cli arguments and passes them to `pipelines.scrinshot_probe_designer` which takes a file path (or file-like binary object) to a config file as its argument. This could then also be used directly by Python scripts like this:

```python
import oligo_designer_toolsuite.pipelines.scrinshot_probe_designer

config_path = "~/Documents/some/config.yaml"
scrinshot_probe_designer(config_file=config_path)
```

In the future, this could also be extended to allow passing a Pydantic config object instead of a file path, which would make programmatic pipeline calls a lot easier (as no temporary config file would be needed anymore).

Regarding command line usage, this setup would also make it easier to pivot to a unified command which calls the appropriate pipeline internally (e.g. `odt run scrinshot -c "~/Documents/some/config.yaml"` would be equivalent to `scrinshot_probe_designer -c "~/Documents/some/config.yaml"`) if that is something that is desired.

Our team would be willing to work on this though we don't consider it urgent.

Tagging @jonas-hag

ersommer:
This PR refactors each probe-design CLI so the full pipeline run lives in a plain Python function that takes an already validated Pydantic model. 
The CLI entry point (main) only parses arguments, loads YAML, validates with Pydantic, and then calls that function.

This allows us to start pipelines from external Python code like this: 
```
import yaml
from pathlib import Path
from oligo_designer_toolsuite.validation import OligoSeqProbeDesignerConfig
from oligo_designer_toolsuite.pipelines import oligo_seq_probe_designer

def _load(filename: str) -> dict:
    with open(Path("data/configs") / filename) as fh:
        data = yaml.safe_load(fh)
        if not isinstance(data, dict):
            raise TypeError
        return data

raw = _load("oligo_seq_probe_designer.yaml")
config = OligoSeqProbeDesignerConfig.model_validate(raw)

oligo_seq_probe_designer(config)
```

**What changed**

- one new function per pipeline, moving the actual execution of the pipeline from main to that new function
- `oligo_designer_toolsuite.pipelines` exports were updated so these runners appear in `__all__` and can be imported from `oligo_designer_toolsuite.pipelines`.

`write_config_to_yaml()` at the moment is called within the pipeline execution function. Potentially it makes sense to move this function call to `main()`, so that it is only executed when the data is coming from a `.yaml `file and not a pydantic model. What do you think about this. 

Since this is based on #149 that should be merged first.

resolves #151


lisa-sousa:
Thanks for the contribution. This PR is now obsolete because we've since reworked the entire Pydantic model. The proposed changes no longer apply to the current implementation, so we're closing this PR.

PR #190, opened 2026-05-13
187 set up basic pydantic infrastructure for oligo seq pipeline

jonas-hag:


oligo_designer_toolsuite/config/_types.py
```diff
@@ -0,0 +1,87 @@
+from typing import Annotated
+
+from pydantic import Field, NonNegativeFloat, NonNegativeInt, PositiveInt
+
```
lisa-sousa, line 4: shoukd we add region_ids as type here as well?

oligo_designer_toolsuite/config/_types.py
```diff
@@ -0,0 +1,87 @@
...
+FilesFastaReferenceDatabaseT = Annotated[
+    FastaFileListT,
+    Field(
+        description="List of paths to FASTA file(s) containing reference sequences against which specificity will be evaluated. These typically include the entire genome or transcriptome to identify off-target binding sites. Hint: use the genomic_region_generator pipeline to create FASTA files of genomic regions of interest."
+    ),
+]
+
+LengthMinT = Annotated[
+    NonNegativeInt, Field(description="Minimum length (in nucleotides) for probe sequences.")
```
lisa-sousa: maybe call it more general here "oligo sequences"

oligo_designer_toolsuite/config/_types.py
```diff
@@ -0,0 +1,87 @@
...
+        description="Maximum GC content (as a fraction between 0.0 and 1.0) for probes. Probes with GC content above this value will be filtered out/rejected.",
+        ge=0,
+        le=100,
+    ),
+]
+GCContentOptT = Annotated[
+    float,
+    Field(
+        description="Optimal GC content for target probes, expressed as a fraction between 0.0 and 1.0. Used in scoring to prioritize probes closer to this value.",
+        ge=0,
```
lisa-sousa, line 60: in the description we say fraction between 0.0 and 1.0 but the ge=0 and le=100, We have to match the sacle everywhere

oligo_designer_toolsuite/config/_types.py
```diff
@@ -0,0 +1,87 @@
...
+        le=100,
+    ),
+]
+
+DNAT = Annotated[str, Field(pattern=r"^[ATGC]+$")]
```
lisa-sousa: we have to think about RNA types as well where we have "U"

oligo_designer_toolsuite/pipelines/_oligo_seq_probe_designer.py
lisa-sousa, line 1054: this function can be removed then no?

oligo_designer_toolsuite/config/_specificity_filters.py
```diff
@@ -0,0 +1,102 @@
...
+]
+
+
+class VariantFilterDisabled(FilterBaseConfig):
+    enabled: Literal[False]
+
+
+class VariantFilterEnabled(FilterBaseConfig):
+    enabled: Literal[True]
+    files_vcf_reference_database: Annotated[
```
lisa-sousa: should this be a type as the blastn reference database?

oligo_designer_toolsuite/config/_property_filters.py
```diff
@@ -0,0 +1,169 @@
...
+
+class TargetedExonsFilterDisabled(FilterBaseConfig):
+    enabled: Literal[False]
+
+
+class TargetedExonsFilterEnabled(FilterBaseConfig):
+    enabled: Literal[True]
+    targeted_exons: list[str] = Field(
+        description="List of exon identifiers that should be preferentially targeted by probes. Only probes in these exons are kept.",
+        default=["1", "2", "3"],
```
lisa-sousa: remove the default=["1", "2", "3"] from the description and add an example into the description

oligo_designer_toolsuite/config/_property_filters.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+class IsoformConsensusFilterEnabled(FilterBaseConfig):
+    enabled: Literal[True]
+    isoform_consensus: Annotated[
+        float,
+        Field(
+            description="Threshold for isoform consensus filtering (typically between 0.0 and 1.0) in %. Probes with isoform consensus values below this threshold will be filtered out. This ensures that selected probes target sequences that are conserved across multiple transcript isoforms.",
+            ge=0,
+            le=100,
```
lisa-sousa, line 38: missmatch between description and "le=100"

oligo_designer_toolsuite/config/_oligo_scoring.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+class GCContentScore(ScoreBaseConfig):
+    GC_content_min: GCContentMinT
+    GC_content_opt: GCContentOptT
+    GC_content_max: GCContentMaxT
+
+
+class TmScore(ScoreBaseConfig):
+    Tm_min: TmMinT
```
lisa-sousa, line 78: weight missing

oligo_designer_toolsuite/config/_oligo_scoring.py
```diff
@@ -0,0 +1,81 @@
...
+
+class TargetedExonsScore(ScoreBaseConfig):
+    targeted_exons: list[str] = Field(
+        description="List of exon identifiers that should be preferentially targeted by probes. Probes overlapping these exons receive higher scores.",
+        default=["1", "2", "3"],
+    )
+
+
+class GCContentScore(ScoreBaseConfig):
+    GC_content_min: GCContentMinT
```
lisa-sousa, line 72: weight missing

oligo_designer_toolsuite/config/_oligo_scoring.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+class IsoformConsensusScore(ScoreBaseConfig):
+    pass
+
+
+class TargetedExonsScore(ScoreBaseConfig):
+    targeted_exons: list[str] = Field(
+        description="List of exon identifiers that should be preferentially targeted by probes. Probes overlapping these exons receive higher scores.",
+        default=["1", "2", "3"],
```
lisa-sousa: remove default=["1", "2", "3"] from description and add example usage

oligo_designer_toolsuite/config/pipelines/oligo_seq_probe_designer.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+class TargetProbeOligoGeneration(BaseModel):
+    model_config = ConfigDict(extra="forbid")
+
+    file_region_ids: str | None = Field(
+        description="file with a list the genes used to generate the probe sequences, leave empty if all the genes are used",
+        default="data/genes/custom_100.txt",
+    )
+    files_fasta_probe_database: FilesFastaDatabaseT = FilesFastaDatabaseT(
```
lisa-sousa: does it make sense to have default values here? Not sure, good for example runs but the user has to define this somehow? Maybe a warning if the default is used?

lisa-sousa (changes requested):

oligo_designer_toolsuite/pipelines/_oligo_seq_probe_designer.py
jonas-hag: I added the validation at the wrong place, the function can stay

oligo_designer_toolsuite/config/_oligo_scoring.py
jonas-hag: is inherited from parent class

oligo_designer_toolsuite/config/_oligo_scoring.py
jonas-hag: is inherited from parent class

oligo_designer_toolsuite/config/_types.py
jonas-hag: where can RNA be used? We can either create a separate RNAT type and allow both or expand DNAT to DRNAT and allow both U and T

oligo_designer_toolsuite/config/_types.py
jonas-hag: good point. I have checked which functions the oligo seq pipeline calls and they expect them in %

oligo_designer_toolsuite/config/_specificity_filters.py
jonas-hag: do we need it in different filters? I have originally started with the FilesFastaReferenceDatabaseT etc. so that we don't have to repeat the description every time we reuse it. If we will reuse the data base for the variants, then it make sense to also have a type, otherwise I think we can keep it as is.

oligo_designer_toolsuite/config/_types.py
jonas-hag: good idea as we will reuse this across pipelines

oligo_designer_toolsuite/config/pipelines/oligo_seq_probe_designer.py
```diff
@@ -0,0 +1,206 @@
...
+        distance_between_target_probes=0,
+        n_attempts_graph=50,
+        n_attempts_clique_enum=50,
+        diversification_fraction=0.1,
+        jaccard_opt=0.5,
+        jaccard_step=0.1,
+    )
+    uniform_distance_score: UniformDistanceScore = UniformDistanceScore(weight=1)
+    isoform_consensus_score: IsoformConsensusScore = IsoformConsensusScore(weight=1)
+    targeted_exons_score: TargetedExonsScore = TargetedExonsScore(weight=0, targeted_exons=["1", "2", "3"])
```
lisa-sousa: can we leave the targeted exons defaults as empyt list here since the weight is also 0. This has not clearly been defined yet fpr the pipeline.

oligo_designer_toolsuite/config/pipelines/oligo_seq_probe_designer.py
```diff
@@ -0,0 +1,206 @@
...
+    LengthMaxT,
+    LengthMinT,
+    RegionListT,
+)
+
+
+class TargetProbeOligoGeneration(BaseModel):
+    model_config = ConfigDict(extra="forbid")
+
+    file_region_ids: RegionListT = "data/genes/custom_100.txt"
```
lisa-sousa: dynamically fill from yaml file:
`file_region_ids: RegionListT | None = None`

oligo_designer_toolsuite/config/pipelines/oligo_seq_probe_designer.py
```diff
@@ -0,0 +1,206 @@
...
+    files_fasta_probe_database: FilesFastaDatabaseT = FilesFastaDatabaseT(
+        [
+            "data/genomic_regions/exon_annotation_source-NCBI_species-Homo_sapiens_annotation_release-110_genome_assemly-GRCh38.fna",
+            "data/genomic_regions/exon_exon_junction_annotation_source-NCBI_species-Homo_sapiens_annotation_release-110_genome_assemly-GRCh38.fna",
+        ]
```
lisa-sousa:
```
files_fasta_probe_database: FilesFastaDatabaseT = Field(
        default_factory=list
    )
```

oligo_designer_toolsuite/config/pipelines/oligo_seq_probe_designer.py
```diff
@@ -0,0 +1,206 @@
...
+        )
+    )
+    specificity_blastn_filter: SpecificityBlastnFilterConfig = SpecificityBlastnFilterEnabled(
+        enabled=True,
+        search_parameters=BlastnSearchParameters(perc_identity=80, strand="minus", word_size=10),
+        hit_parameters=BlastnHitParameters(coverage=50),
+        files_fasta_reference_database=[
+            "data/genomic_regions/exon_annotation_source-NCBI_species-Homo_sapiens_annotation_release-110_genome_assemly-GRCh38.fna",
+            "data/genomic_regions/exon_exon_junction_annotation_source-NCBI_species-Homo_sapiens_annotation_release-110_genome_assemly-GRCh38.fna",
+        ],
```
lisa-sousa: `files_fasta_reference_database=Field(default_factory=list)`

oligo_designer_toolsuite/config/pipelines/oligo_seq_probe_designer.py
```diff
@@ -0,0 +1,206 @@
...
+    )
+    variant_filter: VariantFilterConfig = VariantFilterEnabled(
+        enabled=True,
+        files_vcf_reference_database=["data/annotations/custom_GCF_000001405.40.chr16.vcf"],
+        action="flag",
+    )
```
lisa-sousa, line 128: `files_vcf_reference_database=Field(default_factory=list)`

oligo_designer_toolsuite/config/_general_models.py
```diff
@@ -0,0 +1,512 @@
...
+    n_jobs: PositiveInt = Field(
+        description="number of cores used to run the pipeline and 2*n_jobs +1 of regions that should be stored in cache. If memory consumption of pipeline is too high reduce this number, if a lot of RAM is available increase this number to decrease runtime"
+    )
+    dir_output: str = Field(description="name of the directory where the output files will be written")
+    write_intermediate_steps: bool = Field(
+        description="if true, writes the oligo sequences after each step of the pipeline into a csv file",
+    )
+
+
+class HomopolymerThresholds(BaseModel):
```
lisa-sousa: The name is wrong -> should be HomopolymericRunThreshold

oligo_designer_toolsuite/config/_general_models.py
```diff
@@ -0,0 +1,512 @@
...
+        default=None,
+        validation_alias=AliasChoices("-evalue", "evalue"),
+        serialization_alias="-evalue",
+        description="Expectation value (E) threshold for saving hits. Default = 10 (1000 for blastn-short)",
+    )
+    word_size: NonNegativeInt | None = Field(
+        default=None,
+        validation_alias=AliasChoices("-word_size", "word_size"),
+        serialization_alias="-word_size",
+        ge=4,
```
lisa-sousa, line 75: is tehre no le?

oligo_designer_toolsuite/config/_general_models.py
```diff
@@ -0,0 +1,512 @@
...
+    min_alignment_length: NonNegativeInt | None = Field(
+        default=None,
+        description="Number of nucleotides for alignment, alternatively, coverage can be used",
+    )
+
+    @model_validator(mode="after")
+    def _check_mutually_exclusive(self) -> Self:
+        if (self.coverage is None) == (self.min_alignment_length is None):
+            raise ValueError("Exactly one of 'coverage' or 'min_alignment_length' must be set.")
+        return self
```
lisa-sousa, line 346: In the framework we throw a warning that if both are set we use min_alignment_length. Should we pre-check that here as well? Or even make it mutually exclusive?

oligo_designer_toolsuite/config/_oligo_scoring.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+class ScoreBaseConfig(BaseModel):
+    model_config = ConfigDict(extra="forbid")
+
+    weight: float = Field(description="Weight assigned in the scoring function.")
+
+
+class UniformDistanceScore(ScoreBaseConfig):
+    pass
```
lisa-sousa, line 58: missing implementation?

oligo_designer_toolsuite/config/_oligo_scoring.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+class IsoformConsensusScore(ScoreBaseConfig):
+    pass
```
lisa-sousa, line 62: missing implementation?

oligo_designer_toolsuite/config/_oligo_scoring.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+class TargetedExonsScore(ScoreBaseConfig):
+    targeted_exons: list[str] = Field(
+        description="List of exon identifiers that should be preferentially targeted by probes, e.g. ['1', '2', '3']. Probes overlapping these exons receive higher scores."
+    )
+
```
lisa-sousa, line 69: missing weight parameter?

oligo_designer_toolsuite/config/_oligo_scoring.py
```diff
@@ -0,0 +1,80 @@
...
+
+class GCContentScore(ScoreBaseConfig):
+    GC_content_min: GCContentMinT
+    GC_content_opt: GCContentOptT
+    GC_content_max: GCContentMaxT
+
```
lisa-sousa, line 75: missing weight parameter?

oligo_designer_toolsuite/config/_oligo_scoring.py
```diff
@@ -0,0 +1,80 @@
...
+
+class TmScore(ScoreBaseConfig):
+    Tm_min: TmMinT
+    Tm_opt: TmOptT
+    Tm_max: TmMaxT
```
lisa-sousa, line 80: missing weight parameter?

oligo_designer_toolsuite/config/_property_filters.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+class FilterBaseConfigDisabled(BaseModel):
+    model_config = ConfigDict(extra="ignore")
+    enabled: Literal[False]
+
+
+class IsoformConsensusFilterDisabled(FilterBaseConfigDisabled):
+    pass
+
```
lisa-sousa, line 30: Is there no implementation for Disbaled filter classes?

oligo_designer_toolsuite/config/_types.py
lisa-sousa: I think in scrinshot we use RNA ("U"s) as well.

oligo_designer_toolsuite/config/_property_filters.py
jonas-hag: inherits implementation from `FilterBaseConfigDisabled`

oligo_designer_toolsuite/config/_oligo_scoring.py
jonas-hag: inherits it from `ScoreBaseConfig`

oligo_designer_toolsuite/config/_oligo_scoring.py
jonas-hag: inherits it from `ScoreBaseConfig`

oligo_designer_toolsuite/config/_oligo_scoring.py
jonas-hag: inherits it from `ScoreBaseConfig`

oligo_designer_toolsuite/config/_oligo_scoring.py
jonas-hag: inherits implementation from `ScoreBaseConfig`

oligo_designer_toolsuite/config/_oligo_scoring.py
jonas-hag: inherits it from `ScoreBaseConfig`

oligo_designer_toolsuite/config/_general_models.py
jonas-hag: no, blastn documentation only specifies >=4

oligo_designer_toolsuite/config/_general_models.py
jonas-hag: this function also throws an error when both arguments are set, so I think we already have the necessary checks here. Also, the serialisation removes the value that is set to None, so the function in the framework only receives one argument

oligo_designer_toolsuite/config/_types.py
jonas-hag: I've now expanded it to `DRNAT` that included `U`

jonas-hag:
Introduces a Pydantic v2 config validation layer for the oligo-seq probe designer pipeline, replacing unvalidated dict access with a typed model that is validated at config-load time.

##  New config/ module (_general_models.py, _property_filters.py, _specificity_filters.py, _oligo_scoring.py, _types.py, pipelines/oligo_seq_probe_designer.py)
  - Full Pydantic v2 model hierarchy covering every YAML key for the oligo-seq pipeline
  - Discriminated unions with enabled: true/false for all optional filters (property, specificity, variant, Tm correction)
  - Shared annotated type aliases (FilesFastaReferenceDatabaseT, LengthMinT, …) with field-level descriptions used for JSON-schema / UI generation
  - OligoSeqProbeDesignerConfig exported at the top-level package

##  Pipeline integration (_oligo_seq_probe_designer.py)
  - _load_config now calls OligoSeqProbeDesignerConfig.model_validate() and raises a ValidationError on bad configs
  - Access to Tm_chem/salt_correction_parameters updated to the new nested parameters sub-key; disabled corrections are set to None explicitly before downstream use
  - global → global_parameters throughout (avoids clash with Python built-in)

##  YAML config (data/configs/oligo_seq_probe_designer.yaml)
  - Added schema_version: 2
  - global → global_parameters
  - Tm_chem_correction_parameters and Tm_salt_correction_parameters wrapped with enabled/parameters structure
  - targeted_exons default changed from ["1","2","3"] to []

##  Schema generation support
  - Filters that require user-supplied file paths (specificity_blastn_filter, variant_filter) use model_construct() in their default_factory so a partial default (all params except paths) can be emitted in the JSON schema without triggering a ValidationError

How to export the JSON schema:
```
from oligo_designer_toolsuite.config.pipelines.oligo_seq_probe_designer import OligoSeqProbeDesignerConfig
import json

from typing import Any
from pydantic_core import core_schema                                                                                                                                             
from pydantic.json_schema import GenerateJsonSchema, NoDefault                                                                                                                     

class AllDefaults(GenerateJsonSchema):
    def get_default_value(self, schema: core_schema.WithDefaultSchema) -> Any:
        if "default" in schema:
            return schema["default"]
        elif "default_factory" in schema and not schema.get("default_factory_takes_data"):
            try:
                value = schema["default_factory"]()
            except Exception:
                return NoDefault
            if hasattr(value, "model_dump"):
                return value.model_dump(exclude_unset=True)
            return value
        return NoDefault

with open("schema.json", "w") as f:
    json.dump(OligoSeqProbeDesignerConfig.model_json_schema(schema_generator=AllDefaults), f, indent=4)
```

lisa-sousa (approved):

PR #207, opened 2026-06-30
Minor refactor: Remove iteration_utilities dependency

maxtidev:
### Summary

This PR removes the [iteration_utilities](https://pypi.org/project/iteration-utilities/) dependency. It was only used in `_get_duplicated_sequences` which has been changed to only use the standard library instead. It's a minor refactor and very low-priority.

Changes:
- remove iteration_utilities from `pyproject.toml` and `docs/requirements.txt`
- replace usage of iteration_utilities with standard library

### Reasons for this change

For context, the package uses compiled extensions, i.e. C code that has to be compiled into wheels which are specific per Python version, architecture, etc. This is usually done by the developer of a package who then distributes these wheels on PyPi. However, we came across situations where there was no matching wheel for iteration_utilities. In this scenario, installing ODT via `pip install` automatically downloads the source code for iteration_utilities and compiles it on the user's machine. This means that users need to have a C compiler installed and the installation process takes longer.

We currently install a C toolchain on ARM in our ODT Cloud Docker containers because of this problem (see https://github.com/HelmholtzAI-Consultants-Munich/oligo-designer-toolsuite-front-end/issues/35). [These lines](https://github.com/HelmholtzAI-Consultants-Munich/oligo-designer-toolsuite-front-end/blob/10175c4adc3028bcfe83f2b3a88e4b9d8b7af65a/docker/server.Dockerfile#L15) can likely be removed once ODT does not depend on iteration_utilities anymore.

There are a couple of additional reasons for removing iteration_utilities:
- The newest version is `0.13.0`, implying that it's still considered experimental.
- It does not seem to be very actively maintained ([GitHub repo](https://github.com/MSeifert04/iteration_utilities)).
- ODT uses it in only a single function (replaced in this PR).
- ODT gets more lightweight.
- We won't be limited by iteration_utilities's Python version support anymore (currently <3.14).


lisa-sousa (commented):
The code itself looks good to me, however, we choose iteration_unitlities back then because it was very fast. Did you check how / if the runtime changes by using Counter instead?

maxtidev:
Ah, I see. I wrote a simple micro benchmark to test the initial implementation versus some alternatives:

 ```py
from random import shuffle
import random
from collections import Counter
import time

import iteration_utilities
import pandas as pd

INITIAL_NUMBER_SEQUENCES: int = 1000000
LENGTH_PER_SEQUENCE: int = 40
MIN_DUPLICATE_FRACTION: float = 0.10


def generate_sequences(amount: int, length: int):
    for i in range(amount):
        yield "".join(random.choices(["A", "T", "G", "C", "S"], k=length))


def _duplicates_iteration_utilities(sequences: list) -> list:
    # find the duplicates within the database
    duplicated_sequences = list(
        iteration_utilities.unique_everseen(iteration_utilities.duplicates(sequences))
    )

    return duplicated_sequences


def _duplicates_pandas(sequences: list) -> list:
    seq_series = pd.Series(sequences)
    duplicated_sequences = list(seq_series[seq_series.duplicated()])

    return duplicated_sequences


def _duplicates_counter(sequences: list) -> list:
    # count occurences
    counts = Counter(sequences)

    # find the duplicates within the database
    duplicated_sequences = [seq for seq, count in counts.items() if count > 1]

    return duplicated_sequences


def _duplicates_manual(sequences: list) -> list:
    # keep track
    seen = set()
    duplicated_sequences = []

    # find the duplicates within the database
    for seq in sequences:
        if seq in seen:
            duplicated_sequences.append(seq)
        else:
            seen.add(seq)

    return duplicated_sequences


implementations = [
    _duplicates_iteration_utilities,
    _duplicates_pandas,
    _duplicates_counter,
    _duplicates_manual,
]

print(f"{INITIAL_NUMBER_SEQUENCES=}")
print(f"{LENGTH_PER_SEQUENCE=}")
print(f"{MIN_DUPLICATE_FRACTION=}")

sequences = list(generate_sequences(INITIAL_NUMBER_SEQUENCES, LENGTH_PER_SEQUENCE))

# add duplicates
sequences.extend(sequences[: int(len(sequences) * MIN_DUPLICATE_FRACTION)])
shuffle(sequences)

print("Overall amount of sequences:", len(sequences))
print()

for imp in implementations:
    fun_name = imp.__name__.ljust(32)
    start = time.perf_counter()
    imp(sequences)
    print(fun_name, f"Elapsed time: {time.perf_counter() - start:.3f}s")
```

On my machine, this gives the following output:
```
INITIAL_NUMBER_SEQUENCES=1000000
LENGTH_PER_SEQUENCE=40
MIN_DUPLICATE_FRACTION=0.1
Overall amount of sequences: 1100000

_duplicates_iteration_utilities  Elapsed time: 0.220s
_duplicates_pandas               Elapsed time: 0.325s
_duplicates_counter              Elapsed time: 0.317s
_duplicates_manual               Elapsed time: 0.208s
```
And for a larger number:
```
INITIAL_NUMBER_SEQUENCES=10000000
LENGTH_PER_SEQUENCE=40
MIN_DUPLICATE_FRACTION=0.1
Overall amount of sequences: 11000000

_duplicates_iteration_utilities  Elapsed time: 2.512s
_duplicates_pandas               Elapsed time: 6.396s
_duplicates_counter              Elapsed time: 4.616s
_duplicates_manual               Elapsed time: 2.612s
```
And for a higher amount of actual duplicates:
```
INITIAL_NUMBER_SEQUENCES=1000000
LENGTH_PER_SEQUENCE=40
MIN_DUPLICATE_FRACTION=0.7
Overall amount of sequences: 1700000

_duplicates_iteration_utilities  Elapsed time: 0.361s
_duplicates_pandas               Elapsed time: 0.594s
_duplicates_counter              Elapsed time: 0.493s
_duplicates_manual               Elapsed time: 0.293s
```

While iteration_utilities is indeed faster than the `Counter` approach, it's in the same order of magnitude. The difference shouldn't be noticeable when running an entire pipeline.

I also added the manual Python approach which is similar in speed, if not faster, compared to iteration_utilities. If we want to prevent any sort of performance regression while still benefiting from one less dependency, I'd be happy to change the PR to the manual approach instead. Let me know what you think :)

lisa-sousa:
Thanks for the comparisons!

I would prefer to have the manual python approach then since it we could get rid of iteration_utilities and keep the same speed as before.

lisa-sousa (approved):
Looks good. Thanks for the PR.

PR #214, opened 2026-07-03
move the oligo-seq specific overrides

### LINKED ISSUE #208: Change oligo-seq pydantic structure
We want all parameters that need to be set for one pipeline in one file so that it is clear what needs to be changed for a pipeline.

Move the content of [oligo_seq_probe_designer_overrides.py](https://github.com/HelmholtzAI-Consultants-Munich/oligo-designer-toolsuite/blob/main/oligo_designer_toolsuite/config/overrides/oligo_seq_probe_designer_overrides.py) to [oligo_seq_probe_designer.py](https://github.com/HelmholtzAI-Consultants-Munich/oligo-designer-toolsuite/blob/mainoligo_designer_toolsuite/config/pipelines/oligo_seq_probe_designer.py)

jonas-hag:
move the oligo-seq specific overrides directly in the oligo-seq pipeline file so that all parameters for one pipeline are in the same file

additionally:
- rename `distance_between_target_probes` to `distance_between_probes` to be in line with other pipelines
- add a test for JSON schema generation


jonas-hag:
closes #208

lisa-sousa (approved):
Changes look good.

PR #216, opened 2026-07-08
Add pydantic input validation for the SCRINSHOT pipeline

### LINKED ISSUE #209: Pydantic for SCRINSHOT

jonas-hag:
- Add ScrinshotProbeDesignerConfig with target_probe/detection_oligo models, reusing generic filter/scoring/Tm building blocks and keeping SCRINSHOT-specific overrides (specificity_blastn_filter's ligation_region_size, the distance_between_probes naming) local to config/pipelines/scrinshot_probe_designer.py
- Validate the raw YAML in main() and pass the validated config's model_dump() into scrinshot_probe_designer()
- Export ScrinshotProbeDesignerConfig from oligo_designer_toolsuite.config
- Update TestScrinshotProbeDesigner to derive its output dir from the validated config instead of raw YAML
- Add TestScrinshotYaml covering parsing, round-tripping, extra="forbid" rejection, and JSON schema generation

closes #209


jonas-hag:
@lisa-sousa 
I was not entirely sure if reusing the `TMaxT` etc. types for the padlock arms, I've tried to change the descriptions a bit to make it fit better, but I'm also happy to separate define the fields for the padlock arms if this clarifies the documentation

Also, I've realised that in the YAML in the section of `padlock_arms_properties` we could rename the fields from `padlock_arm_length_min` to `length_min` etc. to have it more concise (as we did it e.g. for the target probes), but not sure if this helps understanding the parameters

oligo_designer_toolsuite/config/_types.py
```diff
@@ -75,7 +75,7 @@
 TmMaxT = Annotated[
     NonNegativeFloat,
     Field(
-        description="Maximum melting temperature (Tm) in degrees Celsius for target probes. Probes with calculated Tm above this value will be filtered out. This value is also used as the optimal Tm target in probe scoring."
+        description="Maximum melting temperature (Tm) in degrees Celsius for probes. Probes with calculated Tm above this value will be filtered out. This value is also used as the optimal Tm target in probe scoring."
     ),
```
lisa-sousa, line 79: @jonas-hag I think it's fine like that. We don't need to specify TmMin / TmMax for each oligo type seperately. It doesn't change the meaning of the parameter itself when applied to different oligo types. But I would remove the last part of the description because it is not used for all oligos in the scoring.

lisa-sousa:
> Also, I've realised that in the YAML in the section of `padlock_arms_properties` we could rename the fields from `padlock_arm_length_min` to `length_min` etc. to have it more concise (as we did it e.g. for the target probes), but not sure if this helps understanding the parameters

I've also thought about it when refactoring th epipeline but somehow I thought it might be too unspecific then. But looking at it now, I think we can remove the padlock_arm prefix from all those padlock_arm_propertiy parameters.

lisa-sousa (approved):

PR #217, opened 2026-07-08
Add pydantic input validation for the HCR pipeline

### LINKED ISSUE #210: Pydantic for HCR

jonas-hag:
- Add HcrProbeDesignerConfig in config/pipelines/hcr_probe_designer.py, keeping HCR-specific overrides local
- Model HCR-specific codebook/initiator_table with source restricted to Literal["load"], rejecting source="generate" at validation since generation is not yet implemented for HCR. It is implemented specific to HCR and not with cylceHCR in mind, because for cycleHCR codebook generation already works
- Validate the raw YAML in main() and pass the validated config's model_dump() through _preprocess_config into hcr_probe_designer()
- Export HcrProbeDesignerConfig from oligo_designer_toolsuite.config
- Update TestHCRProbeDesigner to derive its output dir from the validated config instead of raw YAML
- Add TestHcrYaml covering parsing, round-tripping, extra="forbid" rejection, source="generate" rejection, and JSON schema generation
- additional validation of input files was not implemented yet, this will be done in a future PR

@lisa-sousa please check if the description for `junction_region_size` in the pydantic model is correct.

I suggest to add all input file validation in a separate PR. Also, I've noticed some slight differences in naming in the YAML files. E.g. `target_probe` vs. `initiator_probes`. I think plural makes more sense here, we could probably adjust it in all pipelines together with a PR at the end. Otherwise, I could also change it here.


lisa-sousa:
Yes, let's do the input file validation in the end as a seperate PR. We probably also discover other things along the way that we can fix afetrwrad in a final PR. Regarding the naming convention. I need to do a final PR with some changes over all pipelines. There I would unify the naming.

lisa-sousa:
description for `junction_region_size` is correct.

lisa-sousa:
For the codebook and oligo tables, we have to see at some point if it makes sense to have a general pydantic model for them, if we keep repeating the same thing here.

lisa-sousa (commented):
Looks good. Can be merged when PR 216 is merged.

jonas-hag:
> For the codebook and oligo tables, we have to see at some point if it makes sense to have a general pydantic model for them, if we keep repeating the same thing here.

yes, I think in general this would be possible as there are only a few differences, mostly in the validation. there we need to see how easily it can be done. but only makes sense to look at when load & generate is implemented in both pipelines

lisa-sousa (approved):
Looks all good

PR #226, opened 2026-07-22
221 final adjustments over all pipelines

### LINKED ISSUE #221: Final adjustments over all pipelines
- [x] _utils.py cleanup
  - Drop unused_bits warning in validate_bit_mapping_table; keep missing_bits as FileFormatError.

- [x] Codebook shape invariant across codebook pipelines (SEQFISH+, CycleHCR)
  - Drop codebook column trim in generate_codebook for SEQFISH+ and CycleHCR.
  - Drop readout-probe-table row trim in design_readout_probes for both.
  - Shrink shipped defaults in data/configs/seqfish_plus_probe_designer.yaml (proposed: n_barcode_rounds: 3, n_pseudocolors: 4, keep 3 channels → 36-row readout table).
  - Review CycleHCR shipped defaults; shrink only if readout table would otherwise ship mostly empty.
  - Regenerate example files under data/functional_oligos/ for both pipelines.

- [x] Plural rename + oligo/probe audit
  - Rename YAML top-level key target_probe → target_probes in all six configs.
  - Update all corresponding config["target_probe"] → config["target_probes"] dict accesses in each pipeline module.
  - Audit each pipeline module for oligo vs probe mistakes (keep oligo_database, oligo_generation, oligo_length; keep target_probe_length_min, readout_probe_sequence, etc.); fix only real mistakes.

- [x] Target-probe API symmetry + inner-class canonicalization
  - For each pipeline (order: OligoSeq → SCRINSHOT → HCR → CycleHCR → SEQFISH+ → MERFISH, smoke between each):
    - Add TargetProbeDesigner.generate_target_probes(target_probe_parameters: dict) wrapping the multi-step primitives.
    - Demote create_oligo_database / filter_by_property / filter_by_specificity / create_oligo_sets on TargetProbeDesigner to _-prefixed private helpers.
    - Simplify <Pipeline>ProbeDesigner.design_target_probes to a one-liner calling generate_target_probes.
    - Reorder property + specificity filter build inside _filter_by_property / _filter_by_specificity to the canonical order.
    - Reorder YAML property_filters and specificity_filters sub-block sequence to match the canonical order (property: isoform_consensus → hard_masked → soft_masked → homopolymeric → GC → Tm → secondary_structure; specificity: cross_hybridization → specificity_blastn).
    - Reorder methods inside every inner class (TargetProbeDesigner, ReadoutProbeDesigner, PrimerDesigner, InitiatorDesigner, etc.) to the canonical layout.

- [x] Validate readout probe table before generating codebook
  - For each codebook + readout pipeline (MERFISH, SEQFISH+, CycleHCR):
    - In design_readout_probes, reorder the load/generate branches so the readout probe table is loaded and shape-checked before the codebook is generated (only matters for the load path; harmless for generate).
    - Optional lightweight helper in _utils.py: validate_readout_probe_table_shape(table, n_bits, required_columns) -> None — cheap pre-check (row count, column presence) called before the codebook is generated. Full semantic validation stays in the joint validate at the end.

- [x] Docstring sweep in pipelines/ folder only
  - Sweep every .py file under oligo_designer_toolsuite/pipelines/.
  - For each function, method, and class (public, private, inner, module-level):
    - Confirm a docstring exists; add one if missing.
    - Format: :param name: ... / :type name: ... / :return: ... / :rtype: ... blocks — same style as existing docstrings.
    - Tone: biological rationale in a short intro (e.g., "GC clamp improves primer binding stability" rather than "adds GCClampFilter"), keeping it accessible to non-technical readers.
    - Private helpers get docstrings too, with full :param: blocks — briefer intro is OK.
  - Special attention to methods touched in stages 1–5 (new generate_target_probes, private-helper renames, canonicalized method orders) so their docstrings match the final API shape.
  - Also update _preprocess_config and the <pipeline>_probe_designer(config) orchestrator docstrings to reflect the plural target_probes YAML key + any signature changes from stages 3–4.
  - Also touch _utils.py docstrings (in-scope since it lives under pipelines/).

lisa-sousa:
## Summary
- Align probe-designer YAML configs and pipeline APIs across OligoSeq, SCRINSHOT, HCR, CycleHCR, MERFISH, and seqFISH+ (parameter grouping, `target_probes` / `oligo_database` naming, shared filter order).
- Add shared codebook / bit-mapping validation and loadable example codebook + readout/initiator tables.
- Tighten docs/comments and replace validation asserts with clearer `ConfigurationError` / `ValueError` handling.


oligo_designer_toolsuite/config/pipelines/oligo_seq_probe_designer.py
```diff
@@ -197,4 +197,4 @@ class OligoSeqProbeDesignerConfig(BaseModel):
         write_intermediate_steps=True,
     )
 
-    target_probe: TargetProbe
+    target_probes: TargetProbe
```
jonas-hag: I would then also change the names of the pydantic models to plural, this is in line with the models in the other pipelines

oligo_designer_toolsuite/pipelines/_seqfish_plus_probe_designer.py
```diff
@@ -1386,6 +1400,39 @@ def generate_readout_probe_table(
...
+        (``expected_hamming_weight=None``) because SeqFISH+ uses a combinatorial pseudocolor
+        encoding rather than a fixed-weight binary code.
+        """
+        validate_codebook(
+            codebook=codebook,
+            region_ids=region_ids,
+            source=codebook_source,
+            expected_hamming_weight=None,
+            index_name="gene_name",
+        )
```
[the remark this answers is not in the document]
lisa-sousa: Correct, addressed this below in the new commit

oligo_designer_toolsuite/pipelines/_merfish_probe_designer.py
```diff
@@ -483,33 +482,28 @@ def design_primers(
...
+            pipeline config. This includes the forward and reverse primer entries
+            and their sequences or design settings.
         :type primer_parameters: dict
-        :return: A tuple ``(reverse_primer_sequence, forward_primer_sequence)``.
+        :return: Reverse primer sequence and forward primer sequence.
         :rtype: tuple[str, str]
         """
-        # Write the hybridization probes to a FASTA file so they can be used as a reference for
-        # the second specificity pass (primer must not bind the hybridization probes themselves).
+        # Dump encoding probes so primer design can reject oligos that bind the probe body.
```
jonas-hag: previous comment was more informativ?

oligo_designer_toolsuite/pipelines/_utils.py
```diff
@@ -1,10 +1,20 @@
+"""
+Shared helper functions for probe-design pipelines.
+
+This module contains small utility functions that are used across several
+pipelines in :mod:`oligo_designer_toolsuite.pipelines`.
+
+Keeping these functions in one place avoids repeated code and makes pipeline
```
jonas-hag: too verbose?

oligo_designer_toolsuite/pipelines/_utils.py
```diff
@@ -23,6 +33,16 @@
...
+
+    Each probe-designer command accepts a ``-c / --config`` argument that points
+    to a YAML configuration file. This helper handles that shared argument so
+    individual pipeline ``main()`` functions can stay short and consistent.
+
+    :return: Parsed command-line arguments as a dictionary.
+    :rtype: dict[str, Any]
+    """
     parser = ArgumentParser(
         prog="Genomic Region Generator",
```
jonas-hag: I think prog and usage should be specific to the pipeline; should we make this arguments to base_parser?

oligo_designer_toolsuite/pipelines/_utils.py
```diff
@@ -152,16 +193,24 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
 
 def pipeline_step_advanced(step_name: str) -> Callable[[F], F]:
```
jonas-hag: do we still need this decorator?

oligo_designer_toolsuite/pipelines/_scrinshot_probe_designer.py
```diff
@@ -1451,32 +1513,29 @@ def _find_best_oligo(
...
         :type U_distance: int
-        :return: The detection oligo sequence with T nucleotides converted to U and a fluorophore
-            position indicator added. The indicator "[fluorophore]" is added at the end where
-            fewer T nucleotides are present (to preserve more T nucleotides for conversion).
+        :return: Detection oligo with uracil substitutions and a ``[fluorophore]``
+            marker.
         :rtype: str
         """
-
+        # Attach the dye on the end with fewer nearby T's so UNG sites stay away from the label.
```
jonas-hag: should we define UNG?

oligo_designer_toolsuite/pipelines/_seqfish_plus_probe_designer.py
```diff
@@ -364,167 +345,149 @@ def design_readout_probes(
...
+            and their sequences or design settings.
         :type primer_parameters: dict
-        :return: A tuple ``(reverse_primer_sequence, forward_primer_sequence)``.
+        :return: Reverse primer sequence and forward primer sequence.
         :rtype: tuple[str, str]
         """
-        # Write the hybridization probes to a FASTA file so they can be used as a reference for
-        # the second specificity pass (primer must not bind the hybridization probes themselves).
-        file_fasta_hybridization_probes_database = hybridization_probe_database.write_database_to_fasta(
+        # Dump hybridization probes so primer design can reject primers that anneal to them.
```
jonas-hag: maybe previous comment had more information?

jonas-hag (changes requested):
looks already good, I just have a few small comments.

Besides the comments in the code, I've noticed that the new documentation sometimes misses a bit from the old docs:

- in the methods of the TargetProbe classes, the docs are not as detailed anymore (also a bit in other Classes, but I mostly noticed it there). Where I noticed it most was for _filter_by_specificity which doesn't contain as much information anymore and the language is not as specific anymore as previously (also a bit for other methods such as _create_oligo_database)
- the overviews in the other classes (e.g. Detection Oligo Designer/Readout Probe Designer etc.) are not as detailed anymore (except for cycleHCR and HCR)

just my observations, not sure if we need to change that

oligo_designer_toolsuite/config/pipelines/oligo_seq_probe_designer.py
lisa-sousa: adapted

lisa-sousa:
I reorganized the class and function documentation in general over the pipelines, such that we avoid too many redundancies, which are hard to maintain, because if we change something we need to update the doc stings in multiple class and function descriptions. Now the main pipelines description lives in the main probe designer class and the remaning docstrings are kept short and concise to avoid redundancies.

oligo_designer_toolsuite/pipelines/_utils.py
lisa-sousa: correct, adapted and changed in each pipeline

oligo_designer_toolsuite/pipelines/_utils.py
lisa-sousa: no, rmeoved

oligo_designer_toolsuite/pipelines/_scrinshot_probe_designer.py
lisa-sousa: specified what UNG is

jonas-hag (approved):

PR #229, opened 2026-07-23
feat: integrate BedParser for BED file handling in OligoDatabase and …

### LINKED ISSUE #227: [Genomic Region Generator] Wrong file parsing
## Summary

When computing intergenic regions, two `pd.read_csv` calls use incorrect options for the files they read:

1. Chromosome-length (`.genome`) file
2. `bedtools complement` BED output

Both currently use `comment="t"` and `header=0`, which do not match how these files are produced and can drop or corrupt data.

## Location

`oligo_designer_toolsuite/sequence_generator/_genomic_region_generator.py` in `_compute_intergenic_annotation_strand`:

```python
# empty-gene case: reads annotation.genome
chromosome_length = pd.read_csv(
    file_chromosome_length, sep="\t", comment="t", header=0, names=["seqid", "length"]
)

# gene-present case: reads bedtools complement output
intergenic_annotation = pd.read_csv(
    file_bed_out, sep="\t", comment="t", header=0, names=["seqid", "start_0base", "end"]
)

### PR DESCRIPTION
**Pull request by lisa-sousa**
## Summary
- Add `BedParser` for reading/writing BED files and converting between 0-based and 1-based starts
- Fix intergenic region generation: stop using `comment="t"` / `header=0` on genome and complement BED files (dropped first rows / truncated lines with `t`)
- Fix `write_database_to_bed` to export 0-based BED starts (was writing 1-based FASTA/DB coords)
- Align region-generator temp BED I/O with `BedParser`; expand docstrings; add unit tests

### DIFF
--- a/oligo_designer_toolsuite/sequence_generator/_genomic_region_generator.py
+++ b/oligo_designer_toolsuite/sequence_generator/_genomic_region_generator.py
@@ -233,27 +233,6 @@ class CustomGenomicRegionGenerator:
         :rtype: str
         """
 
-        def _get_chromosome_length() -> str:
-            """
-            Write chromosome lengths from the genome FASTA to a chrom.sizes file.
-
-            The file is used by bedtools when finding regions outside annotated
-            genes. Each line is ``seqid`` and length, tab-separated, with no header.
-
-            :return: Path to the chromosome-length file.
-            :rtype: str
-            """
-            chromosome_lengths = {}
-            for rec in SeqIO.parse(self.sequence_file, "fasta"):
-                chromosome_lengths[rec.id] = len(rec.seq)
-
-            file_chromosome_length = os.path.join(self.dir_output, "annotation.genome")
-            with open(file_chromosome_length, "w") as handle:
-                for key, value in sorted(chromosome_lengths.items()):
-                    handle.write(f"{key}\t{value}\n")
-
-            return file_chromosome_length
-
         def _compute_intergenic_annotation(
             annotation: pd.DataFrame, file_chromosome_length: str
         ) -> pd.DataFrame:
@@ -386,7 +365,7 @@ class CustomGenomicRegionGenerator:
             return intergenic_annotation
 
         # save chromosome sizes as genome file
-        file_chromosome_length = _get_chromosome_length()
+        file_chromosome_length = self._get_chromosome_length()
 
         # get gene annotation entries
         annotation = self._load_annotation()
@@ -1085,6 +1064,27 @@ class CustomGenomicRegionGenerator:
         del annotation
 
         return file_fasta
+
+    def _get_chromosome_length(self) -> str:
+        """
+        Write chromosome lengths from the genome FASTA to a chrom.sizes file.
+
+        The file is used by bedtools when finding regions outside annotated
+        genes. Each line is ``seqid`` and length, tab-separated, with no header.
+
+        :return: Path to the chromosome-length file.
+        :rtype: str
+        """
+        chromosome_lengths = {}
+        for rec in SeqIO.parse(self.sequence_file, "fasta"):
+            chromosome_lengths[rec.id] = len(rec.seq)
+
+        file_chromosome_length = os.path.join(self.dir_output, "annotation.genome")
+        with open(file_chromosome_length, "w") as handle:
+            for key, value in sorted(chromosome_lengths.items()):
+                handle.write(f"{key}\t{value}\n")
+
+        return file_chromosome_length
 
     def _load_annotation(self) -> pd.DataFrame:
         """
--- a/oligo_designer_toolsuite/sequence_generator/_genomic_region_generator.py
+++ b/oligo_designer_toolsuite/sequence_generator/_genomic_region_generator.py
@@ -55,11 +55,17 @@ class CustomGenomicRegionGenerator:
     labels. Region extractors (gene, exon, intron, CDS, UTR, intergenic, exon-exon junction)
     write one FASTA file per region type.
 
-    Each sequence header starts with ``>`` and holds region ID, optional metadata, and
-    coordinates (chromosome, start, end, strand). Coordinates in the header are 1-based.
-    The region ID is required; other fields are optional.
+    Coordinate indexing:
 
-    Output Format (per sequence)::
+    Annotation files (GFF/GTF) use 1-based starts. Sequence extraction writes temporary
+    BED files, which use 0-based starts. FASTA headers written by this class also use
+    1-based starts so they stay aligned with the annotation.
+
+    Each sequence header starts with ``>`` and holds region ID, optional metadata, and
+    coordinates (chromosome, start, end, strand). The region ID is required; other
+    fields are optional.
+
+    Output Format (per sequence):
 
         >{region_id}::{additional information}::{chromosome}:{start}-{end}({strand})
         sequence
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: oligo_designer_toolsuite/utils/_sequence_parser.py [hunk 1]
```diff
@@ -657,3 +759,134 @@ def merge_vcf_files(self, files_in: list[str], file_out: str) -> None:
...
+            skiprows = [i for i, line in enumerate(handle) if not self._is_bed_data_line(line)]
+
+        return pd.read_csv(
+            filepath,
+            sep="\t",
+            header=None,
+            names=names,
+            skiprows=skiprows,
+            **kwargs,
+        )
```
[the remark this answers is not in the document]
lisa-sousa: addressed in push below

lisa-sousa:
Additional edit in this PR to fix a bug that made Python 3.12 tests fail:
- added region_id parameter to property_filter to allow calculating properties only for one selected region
- solved issues when calculatin gthe seed region site for BlastN filter where we would have concurrent access to the oligo database, i.e. two prallelization loops over the same oligo databse with inplace deletion causing key errors in th eother loop

oligo_designer_toolsuite/sequence_generator/_genomic_region_generator.py
```diff
@@ -273,28 +300,29 @@ def _compute_intergenic_annotation_strand(
             seqid: str, gene_annotatio: pd.DataFrame, strand: str, file_chromosome_length: str
```
jonas-hag: should we fix typo gene_annotatio to gene_annotation as well?

oligo_designer_toolsuite/sequence_generator/_genomic_region_generator.py
```diff
@@ -460,37 +492,41 @@ def get_sequence_exon(self, collapse_duplicated_regions: bool = True) -> str:
...
+        numbers. Coordinates in the FASTA header are 1-based; BED used for
+        extraction uses 0-based starts.
 
-        Output Format (per sequence):
-        >{gene_id}::source={source};species={species};annotation_release={annotation_release};
-        genome_assembly={genome_assembly};regiontype={regiontype};gene_id={gene_id};transcript_id={transcript_id_a},
-        intron_number={intron_number_x};transcript_id={transcript_id_b},intron_number={intron_number_y};
-        number_total_transcripts={number_total_transcripts}::{chromosome}:{start}-{end}({strand})
-        sequence
+        Output Format (per sequence)::
```
jonas-hag: one colon too many? see also subsequent examples

oligo_designer_toolsuite/sequence_generator/_genomic_region_generator.py
```diff
@@ -38,35 +49,39 @@
 
 class CustomGenomicRegionGenerator:
```
jonas-hag, line 50: I'm not sure if the documentation is optimal yet. It makes sense to mention 0/1-based indexing here for the overview (maybe even expand this a little bit) and in functions that are concerned with it (e.g. _load_annotation). However, I'm hot sure if it makes sense to have it in the docstrings for get_sequence_* because there you can't directly do anything with it/no direct arguments for annnotations. Maybe it makes more sense as comments directly in the code where 0/1-based indexing is assumed? But not sure what's the best way here, open to suggestions. Definitely important to document it properly

tests/test_utils.py
```diff
@@ -418,6 +420,123 @@ def test_load_annotation_from_pickle_file(self) -> None:
...
+
+        self.parser.write_bed(
+            bed_in,
+            file_bed,
+            columns=["seqid", "start", "end", "name", "score", "strand"],
+        )
+
+        with open(file_bed) as handle:
+            first_line = handle.readline()
+        assert not first_line.startswith(("track", "browser", "#"))
```
jonas-hag, line 483: not sure if this is needed as we don't add this in write_bed to bed files?

tests/test_utils.py
```diff
@@ -418,6 +420,123 @@ def test_load_annotation_from_pickle_file(self) -> None:
...
+        assert self.parser.convert_start(50, "0-based", "0-based") == 50
+        assert self.parser.convert_start(50, "1-based", "1-based") == 50
+
+    def test_convert_start_invalid_indexing_raises(self) -> None:
+        with self.assertRaises(ConfigurationError):
+            self.parser.convert_start(10, "zero", "1-based")
+        with self.assertRaises(ConfigurationError):
+            self.parser.convert_start(10, "0-based", "one")
+
+    def test_chromosome_length_file_keeps_first_row_with_header_none(self) -> None:
```
jonas-hag: I think we need to test bed_parser.read_bed here as well instead of pd.read_csv

oligo_designer_toolsuite/utils/_sequence_parser.py
```diff
@@ -657,3 +759,142 @@ def merge_vcf_files(self, files_in: list[str], file_out: str) -> None:
...
+        :param columns: Columns to write, in order. If ``None``, all columns are written.
+        :type columns: list[str] | None
+        :return: None
+        """
+        bed_to_write = bed[columns] if columns is not None else bed
+        bed_to_write.to_csv(filepath, sep="\t", header=False, index=False)
+
+    def convert_start(
+        self,
+        start_coordinates: Any,
```
jonas-hag: should we make this narrower with pd.Series? I think so far we only use it with pandas columns and not single values; lists don't work anyway

jonas-hag (changes requested):
looks already really good; I've added a few suggestions

tests/test_utils.py
lisa-sousa: No, we when we read bed files from NCBI/Ensemble, they will have comment lines

oligo_designer_toolsuite/sequence_generator/_genomic_region_generator.py
lisa-sousa: I adapted it, such that we have more code commenst where relevant, and a better explanation in the class description and more narrow explanation in the function description. I still kept a little reference tehre, because I think even though maybe don't modify anything there it's good to know if we use 0 or 1-based indexing tehre.

oligo_designer_toolsuite/utils/_sequence_parser.py
lisa-sousa: I changed to pd.Series

tests/test_utils.py
lisa-sousa: This test covers chrom.sizes files (seqid + length), not BED intervals. The corresponding code snippet still loads those with pd.read_csv, so the test mirrors that approach. read_bed is for BED (track/browser/# handling); intergenic BED reading is already covered by the other TestBedParser cases.

jonas-hag: I see, sorry I mixed it up with a different code snippet.
I still think that this test doesn't make sense because the actual code of `_compute_intergenic_annotation_strand` and the test could diverge. we should either remove the test or have a helper function that reads in the chrom.size files that we can then test

jonas-hag (changes requested):
looks really good! the only issue is the test

tests/test_utils.py
lisa-sousa: yes, gree. This test was developed to implement the correct pd.read_csv loading but doesn't make sense as is. I changed it now to test the generation of the chromosome sizes file and moved the test to the respective GenomicRegionGenerator test class. Therefore, I had to move the nested function to a private helper function.

@jonas-hag please check if you thinks it's ok now like that.

jonas-hag: TBH I'm not sure if the problem is solved as we still directly test pd.read_csv while the implementation in _compute_intergenic_annotation_strand could diverge. I would keep the previous test but add

```
def _read_chromosome_length_file(file_chromosome_length: str):
   return chromosome_length = pd.read_csv(
                    file_chromosome_length, sep="\t", header=None, names=["seqid", "length"]
                )
```
as a method to CustomGenomicRegionGenerator and use it in _compute_intergenic_annotation_strand

then we can also test _read_chromosome_length_file

lisa-sousa: If we want to test that the file is read with the correct parameters, yes, we can encapsulate the read function. I changed the code accordingly.

jonas-hag (approved):
Looks good!

PR #234, opened 2026-08-10
212 pydantic for seqfish

### LINKED ISSUE #212: Pydantic for seqFISH+

jonas-hag:
add a pydantic model for the seqfish pipeline

What is not done yet in this PR:
- more involved input validation (codebooks etc.)
- unified models for codebooks/readout probe tables across pipelines

### DIFF
--- a/oligo_designer_toolsuite/config/_oligo_scoring.py
+++ b/oligo_designer_toolsuite/config/_oligo_scoring.py
@@ -20,12 +20,15 @@ class IndependentSetSelection(BaseModel):
 
     n_sets: PositiveInt = Field(
         description="Number of probe sets to generate per gene.",
+        json_schema_extra={"x-quick-setting": True},
     )
     set_size_min: PositiveInt = Field(
         description="Minimum set size. Genes that cannot form a set of at least this size are reported as insufficient.",
+        json_schema_extra={"x-quick-setting": True},
     )
     set_size_opt: PositiveInt = Field(
         description="Preferred (optimal) number of oligos per set.",
+        json_schema_extra={"x-quick-setting": True},
     )
     distance_between_probes: int = Field(
         description="Required spacing: negative = allow overlap (bases), 0 = adjacent OK, positive = minimum gap (bases).",
--- a/oligo_designer_toolsuite/config/_oligo_scoring.py
+++ b/oligo_designer_toolsuite/config/_oligo_scoring.py
@@ -94,14 +94,14 @@ class UTRScore(BaseModel):
     )
 
 
-class GCContentScoreBasic(BaseModel):
+class GCContentScore(BaseModel):
     model_config = ConfigDict(extra="forbid")
 
     weight: float = Field(description="Weight for GC content in the set-selection scoring function.")
     GC_content_opt: GCContentOptT
 
 
-class GCContentScoreExpanded(GCContentScoreBasic):
+class GCContentScoreNormalized(GCContentScore):
     GC_content_min: GCContentMinT
     GC_content_max: GCContentMaxT
 
--- a/oligo_designer_toolsuite/config/_oligo_scoring.py
+++ b/oligo_designer_toolsuite/config/_oligo_scoring.py
@@ -10,6 +10,10 @@ from oligo_designer_toolsuite.config._types import (
     TmOptT,
 )
 
+PROBE_SET_SELECTION_DESC = (
+    "Parameters for selecting multiple probe sets per gene (size, spacing, scoring, diversification)."
+)
+
 
 class IndependentSetSelection(BaseModel):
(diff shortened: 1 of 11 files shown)


oligo_designer_toolsuite/config/_oligo_scoring.py
```diff
@@ -63,23 +63,46 @@ class ScoreBaseConfig(BaseModel):
...
+
+
+class GCContentScoreBasic(BaseModel):
+    model_config = ConfigDict(extra="forbid")
+
+    weight: float = Field(description="Weight for GC content in the set-selection scoring function.")
     GC_content_opt: GCContentOptT
+
+
+class GCContentScoreExpanded(GCContentScoreBasic):
```
lisa-sousa: This relates to the `NormalizedDeviationFromOptimalGCContentScorer` no? Could we accordinly call it `GCContentScoreNormalized` or simialr such that we know to which scoring function it belongs to?

oligo_designer_toolsuite/config/pipelines/seqfish_plus_probe_designer.py
```diff
@@ -0,0 +1,546 @@
...
+            )
+        return self
+
+
+class TargetProbePropertyFilter(BaseModel):
+    model_config = ConfigDict(extra="forbid")
+
+    isoform_consensus_filter: IsoformConsensusFilterConfig = Field(
+        default=IsoformConsensusFilterEnabled(enabled=True, isoform_consensus=100),
+        description="Require oligos to be supported by a minimum fraction of gene transcripts.",
```
lisa-sousa: shouldn#T those descriptions be part of the general models? If we repeat the description in each pipeline it's quite a maintanance overhead no?

oligo_designer_toolsuite/config/pipelines/seqfish_plus_probe_designer.py
```diff
@@ -0,0 +1,546 @@
...
+# seqFISH+ default search/hit parameters; the path stays required, so the whole
+# filter must be supplied by the user.
+
+
+class SeqfishPlusSpecificityBlastnFilterDisabled(SpecificityBlastnFilterDisabled):
+    pass
+
+
+class SeqfishPlusTargetSpecificityBlastnFilterEnabled(SpecificityBlastnFilterEnabled):
+    search_parameters: BlastnSearchParameters = Field(
```
lisa-sousa: why do we have Blast default parameters here but for the cross-hybridization we define them in the seperate classes of target and readout probes?

lisa-sousa (changes requested):
Looks good, see comments for the seqfish pydantic model.

oligo_designer_toolsuite/config/_oligo_scoring.py
jonas-hag: yes, I wanted to keep this separate so that in case we change something in the pipeline we don't need to rename it in the pydantic model; I will change it

oligo_designer_toolsuite/config/pipelines/seqfish_plus_probe_designer.py
jonas-hag: we need the override for the specificity BlastnFilter parameters because it also expects the file paths to the references which we want to leave empty. to get a valid pydantic model, we need the override here. For the cross-hybridization filters, we don't need to specify the file paths, so we can directly specify the default Blastn parameters in the respective models for Target Probes/readout probes

oligo_designer_toolsuite/config/pipelines/seqfish_plus_probe_designer.py
jonas-hag: I've changed the descriptions so that they are at the most reusable places:
- for property filters with discriminated unions, they now live as `description` in the `Field` of the discriminated union
- for property filters that are just pydantic models I've added docstrings to the models -> should be also parsed by react-jsonschema-forms
- some descriptions are repeated across pipelines, but the models are defined for every pipeline separately. For these, the descriptions are stored in consts

oligo_designer_toolsuite/config/pipelines/seqfish_plus_probe_designer.py
lisa-sousa: Ah ok, so this will probably disappear once we seperate the input files to a seperate pydantic section, right?

lisa-sousa (approved):
Looks good to me.

oligo_designer_toolsuite/config/pipelines/seqfish_plus_probe_designer.py
jonas-hag: yes, then we should be able to do it the same way as for the cross hybridization filters

PR #235, opened 2026-08-10
Fix: Only keep one of each duplicate in `_get_duplicated_sequences`

maxtidev:
This PR is a fix for a regression introduced in #207, see https://github.com/HelmholtzAI-Consultants-Munich/oligo-designer-toolsuite/pull/233/changes#r3748174315.

Changes:
- add a sentence on >2 occurrences duplicate handling to the docstring
- replace `duplicated_sequences = []` with `duplicated_sequences = set()`
- replace `return duplicated_sequences` with `return list(duplicated_sequences)`

There's a performance benefit to be had downstream when keeping the returned duplicates as a set, but that would be a larger change that requires careful review. This is just the fix.


lisa-sousa (approved):
Loos good

PR #236, opened 2026-08-13
213 pydantic for merfish

jonas-hag:
add a pydantic model for the merfish pipeline

What is not done yet in this PR:

- more involved input validation (codebooks etc.)
- unified models for codebooks/readout probe tables across pipelines
- quick setting schema markers for paramters that depend on `source` == "load"/"generate" if they are needed -> we need to figure out best way here first


oligo_designer_toolsuite/config/__init__.py
```diff
@@ -2,10 +2,14 @@
...
+from oligo_designer_toolsuite.config.pipelines.seqfish_plus_probe_designer import (
+    SeqfishPlusProbeDesignerConfig,
+)
 
 __all__ = [
     "CycleHcrProbeDesignerConfig",
     "HcrProbeDesignerConfig",
     "OligoSeqProbeDesignerConfig",
     "ScrinshotProbeDesignerConfig",
+    "SeqfishPlusProbeDesignerConfig",
```
lisa-sousa, line 16: merfish is missing?

lisa-sousa (changes requested):
Looks generally good. I think you're just missing one merfish import of th epydantic model

oligo_designer_toolsuite/config/__init__.py
jonas-hag: yes, I've added it

PR #237, opened 2026-08-25
164 infrastructure change from conda to docker

jonas-hag:
- remove the hybridization filter and the oligo-designer-toolsuite-AI filter dependency
- add a Dockerfile based on micromamba as the bioinformatics dependencies are the easiest to install this way
- describe docker container in README
- add GH actions that automatically adds a new container image to ghcr.io/helmholtz-ai-consultants-munich/odt when a new git tag is added (with the exact version and also latest)

I couldn't test the GH actions yet, we have to see when we create a new release if everything works and make the container registry public.


jonas-hag:
why do we use `types: [created]` in the release.yaml workflow? this also executes when a draft release is made. Is this intentional? Otherwise, we could switch to `published` to be more stringent

lisa-sousa (approved):
Looks all good

PR #238, opened 2026-08-26
correct BLASTN hit criteria and simplify pipeline configs for the front end

yarkinerenn:
## Changes

- **Fixed BLASTN hit criteria being read as the wrong unit.** The criterion used to be set by a
  default on the filter config above `hit_parameters`. Form builders only look at a default on the
  field itself when picking a union branch, so they picked the first one, `coverage`, but the outer
  `value` still came through. In merfish and seqfish that turned `15` (nucleotides) into 15 %
  coverage.
- **Added coverage and min-alignment versions of the BLASTN filters** in `_specificity_filters.py`,
  so the default sits on `hit_parameters` where it gets read. Each pipeline now picks the one it
  wants. The coverage versions are there too, otherwise the rest only work while `coverage` happens
  to be first in the union.
- **Split a `...ProbeDesignerConfigBase` out of all six pipeline configs**, holding everything
  except `general` and `required_parameters`. The front end builds its form from the base.
  `general` is set by the server anyway and was being stripped later on. `...ProbeDesignerConfig`
  still validates the same configs as before.
- **Marked 17 more fields as `x-quick-setting`**: probe length limits in every pipeline, plus
  scrinshot's `min_thymines`, `U_distance` and oligo length, and the same fields in hcr and
  cyclehcr. ODT-Cloud used to keep this list by hand and it had to be updated whenever a field moved.

### DIFF
--- a/oligo_designer_toolsuite/config/_general_models.py
+++ b/oligo_designer_toolsuite/config/_general_models.py
@@ -20,6 +20,27 @@ from oligo_designer_toolsuite.config._types import (
 )
 
 
+# High-level descriptions of the config sections, shown as the section's help text in the form.
+REQUIRED_PARAMETERS_DESC = (
+    "Parameters required for target probe generation. The reference genome is reused for all "
+    "specificity filters for all probes."
+)
+OLIGO_GENERATION_DESC = "Parameters that determine length and spacing of the probes."
+PROPERTY_FILTERS_DESC = (
+    "Parameters for filters that depend on properties of the filters (e.g., melting temperature)."
+)
+SPECIFICITY_FILTERS_DESC = "Parameters for filters that test the specificity of the probes."
+GLOBAL_PARAMETERS_DESC = (
+    "Parameters for melting temperature computation that are reused for different filtering and "
+    "scoring methods."
+)
+CODEBOOK_DESC = "Parameters that determine the codebook generation or loading."
+READOUT_PROBE_TABLE_DESC = "Parameters that determine the readout probe table generation or loading."
+INITIATOR_TABLE_DESC = "Parameters that determine the initiator table generation or loading."
+FORWARD_PRIMER_DESC = "Parameters that determine the forward primer generation or loading."
+REVERSE_PRIMER_DESC = "Parameters that determine the reverse primer generation or loading."
+
+
 class General(BaseModel):
     model_config = ConfigDict(extra="forbid")
     n_jobs: PositiveInt = Field(
--- a/oligo_designer_toolsuite/config/_general_models.py
+++ b/oligo_designer_toolsuite/config/_general_models.py
@@ -30,6 +30,9 @@ PROPERTY_FILTERS_DESC = (
     "Parameters for filters that depend on properties of the filters (e.g., melting temperature)."
 )
 SPECIFICITY_FILTERS_DESC = "Parameters for filters that test the specificity of the probes."
+PROBE_SET_SELECTION_DESC = (
+    "Parameters for selecting multiple probe sets per gene (size, spacing, scoring, diversification)."
+)
 GLOBAL_PARAMETERS_DESC = (
     "Parameters for melting temperature computation that are reused for different filtering and "
     "scoring methods."
--- a/oligo_designer_toolsuite/config/_specificity_filters.py
+++ b/oligo_designer_toolsuite/config/_specificity_filters.py
@@ -2,12 +2,19 @@ from typing import Annotated, Literal
 
 from pydantic import BaseModel, ConfigDict, Field, NonNegativeInt
 
-from oligo_designer_toolsuite.config._general_models import BlastnHitParameters, BlastnSearchParameters
+from oligo_designer_toolsuite.config._general_models import (
+    BlastnHitParameters,
+    BlastnHitParametersCoverage,
+    BlastnHitParametersMinAlignmentLength,
+    BlastnSearchParameters,
+)
 from oligo_designer_toolsuite.config._types import VCFReferenceDatabaseT
 
 SPECIFICITY_TARGET_DESC = "Ensure oligo specificity against a reference database (BLAST-based)."
 SPECIFICITY_READOUT_DESC = "Remove readout probes with significant hits to the reference (BLAST-based)."
 SPECIFICITY_PRIMER_DESC = "Ensure primer specificity against a reference database (BLAST-based)."
+CROSS_HYBRIDIZATION_DESC = "Remove oligos that may cross-hybridize to other probes (BLAST-based)."
+HYBRIDIZATION_PROBES_DESC = "Remove primers that match the assembled hybridization probes (BLAST-based)."
 
 
 class FilterBaseConfigEnabled(BaseModel):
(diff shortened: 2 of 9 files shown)


oligo_designer_toolsuite/config/_general_models.py
```diff
@@ -20,6 +20,27 @@
 )
 
 
+# High-level descriptions of the config sections, shown as the section's help text in the form.
+REQUIRED_PARAMETERS_DESC = (
```
jonas-hag, line 24: let's also move PROBE_SET_SELECTION_DESC to _general_models.py, then all descriptions are in one place

oligo_designer_toolsuite/config/_specificity_filters.py
```diff
@@ -132,3 +139,80 @@ class VariantFilterEnabled(FilterBaseConfigEnabled):
 
 
 VariantFilterConfig = Annotated[VariantFilterEnabled | VariantFilterDisabled, Field(discriminator="enabled")]
+
+
```
jonas-hag, line 153: please reformulate this comment. It needs more context from the frontend and a clearer description

oligo_designer_toolsuite/config/_specificity_filters.py
```diff
@@ -132,3 +139,80 @@ class VariantFilterEnabled(FilterBaseConfigEnabled):
...
+# coverage stays declared first.
+#
+# Values are placeholders, overridden per call site. `description` is restated because
+# redeclaring a field replaces its whole FieldInfo.
+
+
+class SpecificityBlastnFilterMinAlignmentEnabled(SpecificityBlastnFilterEnabled):
+    hit_parameters: BlastnHitParameters = Field(
+        default=BlastnHitParametersMinAlignmentLength(value=15),
+        description=SpecificityBlastnFilterEnabled.model_fields["hit_parameters"].description,
```
jonas-hag: for simplicity, I would also use consts with the description and reuse this across the models that needs them; similar to `description=CROSS_HYBRIDIZATION_DESC`

oligo_designer_toolsuite/config/pipelines/cycle_hcr_probe_designer.py
```diff
@@ -64,14 +70,18 @@ class CycleHcrSpecificityBlastnFilterEnabled(SpecificityBlastnFilterEnabled):
...
     )
-    search_parameters: BlastnSearchParameters = BlastnSearchParameters(
-        perc_identity=100,
-        strand="minus",
-        word_size=10,
-        dust="no",
-        soft_masking=False,
-        max_target_seqs=10,
-        max_hsps=1000,
+    search_parameters: BlastnSearchParameters = Field(
```
jonas-hag, line 75: is this change necessary? in my tests it was still collapsed when using the previous version; but maybe I set up the docker containers wrongly

oligo_designer_toolsuite/config/pipelines/hcr_probe_designer.py
```diff
@@ -63,14 +71,18 @@ class HcrSpecificityBlastnFilterEnabled(SpecificityBlastnFilterEnabled):
...
     )
-    search_parameters: BlastnSearchParameters = BlastnSearchParameters(
-        perc_identity=100,
-        strand="minus",
-        word_size=10,
-        dust="no",
-        soft_masking=False,
-        max_target_seqs=10,
-        max_hsps=1000,
+    search_parameters: BlastnSearchParameters = Field(
```
jonas-hag, line 76: same as in cylcleHCR: is the necessary here?

oligo_designer_toolsuite/config/pipelines/hcr_probe_designer.py
```diff
@@ -81,7 +93,7 @@ class HcrSpecificityBlastnFilterEnabled(SpecificityBlastnFilterEnabled):
 ]
 
 
-############################################
+############################################ add here expanding toggle
```
jonas-hag, line 101: stray comment

oligo_designer_toolsuite/config/pipelines/scrinshot_probe_designer.py
```diff
@@ -58,20 +64,27 @@
...
 class ScrinshotSpecificityBlastnFilterEnabled(SpecificityBlastnFilterEnabled):
-    search_parameters: BlastnSearchParameters = BlastnSearchParameters(
-        perc_identity=80,
-        strand="minus",
-        word_size=10,
-        dust="no",
-        soft_masking=False,
-        max_target_seqs=10,
-        max_hsps=1000,
+    search_parameters: BlastnSearchParameters = Field(
```
jonas-hag, line 78: is this needed? see cycleHCR

jonas-hag (changes requested):

oligo_designer_toolsuite/config/pipelines/cycle_hcr_probe_designer.py
yarkinerenn: I tested one more time, and it is necessary because when we are redefining the defaults, the other Field information on the parents is lost, so we have to redefine. Your experience could be from stale containers

oligo_designer_toolsuite/config/pipelines/hcr_probe_designer.py
yarkinerenn: I tested one more time, and it is necessary because when we are redefining the defaults, the other Field information on the parents is lost, so we have to redefine. Your experience could be from stale containers

oligo_designer_toolsuite/config/pipelines/scrinshot_probe_designer.py
yarkinerenn: I tested one more time, and it is necessary because when we are redefining the defaults, the other Field information on the parents is lost, so we have to redefine. Your experience could be from stale containers

oligo_designer_toolsuite/config/_specificity_filters.py
```diff
@@ -141,22 +151,26 @@ class VariantFilterEnabled(FilterBaseConfigEnabled):
...
+# Variants of the filters above that set the default hit criterion (coverage vs. minimum
+# alignment length) on the `hit_parameters` field itself, not on an enclosing config class.
 #
-# A JSON Schema consumer picks the branch from a default on the field; one further out is not
-# read, so it falls back to the first branch while the outer value still merges in. That made
-# `value=15` meant as 15 bases arrive as 15 % coverage -- silently, both being plain numbers.
-# Coverage variants are needed for the same reason: otherwise a pipeline is correct only while
-# coverage stays declared first.
+# The ODT-Cloud form (react-jsonschema-form) preselects a union branch from the default on the
+# field itself; a default declared further out is ignored for that choice, though its values
```
jonas-hag: thanks, the comment is already way better! I still find this part a little bit confusing: "a default declared further out is ignored for that choice, though its values still merge in"

I think it would help if you could supply a concrete example here (with the different models and where the defaults are taken from), then it should be clear

jonas-hag (changes requested):
thanks, looks good! I would just expand the comment even further

PR #248, opened 2026-09-10
correct genomic region generator YAML for NCBI

jonas-hag:
Currently, there is a bug how the parsed YAML is used for the NCBI genomic region generator. In order to work, unused parameters from the `mode` option that is not used need to be set to `null` in the YAML.

I use this temporary fix as the changed pipelines validate the YAML with pydantic in a future release.


lisa-sousa (approved):
Looks good for now.

PR #249, opened 2026-09-10
Release: merge dev into main

jonas-hag:
Create new release that includes the following functionality:

- remove hybridization filter -> this removes the dependency on pytorch and makes the install more lightweight
- add GH actions to create a docker container for a new release
- fix bug in NCBI genomic region generator so that it works with the shipped YAML and a hint when a different mode is used (short term fix -> YAML validation is changing soon with pydantic validation anyway)


lisa-sousa (approved):
Looks all good

